PR #174, opened 2015-06-06
Converted install guides to Markdown for easier online viewing

jarthorn:
For easier viewing online, and for those who don't have MS Word, it is easier to have install guides in Markdown format. I have just converted them to markdown with no further changes.


mleblanc321:
yup ... docs there for now are just first-pass versions ... (we'll work on 'em); thx

chantisnake:
I will add that to the document page, and I think PDF is the best viewing format (Because I don't have any office on my Linux Machine either)

chantisnake:
install guide is already online. despite its intense ugliness, it is totally functional.

http://lexodocument.github.io/for%20user/2015/06/07/windows-install-guide/
http://lexodocument.github.io/for%20user/2015/06/07/mac-osx-install-guide/

I find that the linux install guide is exactly the same as Mac OS, including the title in the file, so there is no linux version of the install guide. If you think this is okay, then you can close this issue.

I am working on the install Guide with screen shoot.

PR #499, opened 2017-01-22
Python3 migration 

chantisnake:
the result of python3 migration

I have tested with `dan` and get good results on every page (besides stats, but I think that is not releated to python3, so I will fix it after it is merged)

----
To test, simply use 
```
cd Lexos
git checkout py3
python3 lexos.py
```


chantisnake:
If all of you guys think this is okay change, I will merge this to master

scottkleinman:
Good job! I think we'll need to test this a little bit more, and we need to continue to address some bugs in Lexos 3.0, which then have to be pushed to the server. So it's probably not a good idea to merge with the master branch just yet. If you're itching for another task, next week, after my Wednesday class, I'll produce a road map.

chantisnake:
I don't know that, I thought 3.0 is already on the server...

scottkleinman:
That is correct. We anticipate doing bug fixes and updating Lexos 3.0 on the server as necessary before we move to full-time development in Python 3. In other words, there will be Lexos 3.1 (or some increment), and we'll probably use Python 3 on the roadmap to Lexos 4.0.

chantisnake:
I think since we are slowing things down, this release probably will not be 4.0. (maybe 3.5?)

scottkleinman:
Well, yes. When I do the road map, we can tweak it to whatever we agree is appropriate.

chantisnake:
closing this pr for now.

PR #546, opened 2017-07-20
Feature/cut by char test

Weiqi97:
Description
- Test empty string
- Test chunk size parameter
- Test overlap parameter
- Test last proportion parameter
- Test all functionalities together


lexos/helpers/error_messages.py
```diff
@@ -0,0 +1,3 @@
+OVERLAP_SIZE_ERROR = "Overlap has to be smaller than segment size"
+NON_NEGATIVE_ERROR = "Parameter has to be bigger than or equal to zero"
+LAGER_THAN_ZERO_ERROR = "Parameter has to be bigger than zero"
```
chantisnake: use `ERROR_MESSAGE` or simply `MESSAGE` will be better here.

test/unit_test/test_cut_char.py
```diff
@@ -0,0 +1,67 @@
...
+        assert cut_by_characters(text="WORD", chunk_size=2, overlap=0,
+                                 last_prop=0) == ["WO", "RD"]
+        assert cut_by_characters(text="ABBA", chunk_size=2, overlap=1,
+                                 last_prop=0) == ["AB", "BB", "BA"]
+        assert cut_by_characters(text="ABCDE", chunk_size=3, overlap=2,
+                                 last_prop=0) == ["ABC", "BCD", "CDE"]
+        assert cut_by_characters(text="ABCDEF", chunk_size=4, overlap=3,
+                                 last_prop=0) == ["ABCD", "BCDE", "CDEF"]
+
+    def test_string_last_prop(self):
```
chantisnake: would you please add a test so that the `last_prop` is larger than 1?

PR #548, opened 2017-07-20
feature/xinru#10

XinruLiu:


lexos/processors/prepare/cutter.py
```diff
@@ -407,18 +407,17 @@ def cut_by_number(text, num_chunks):
...
-    Returns: A list of strings which are to become the new chunks.
+    Chunk boundaries should be created when every milestone appears
+    :param text: the text to be chunked as a single string
+    :param cutting_value: # maybe we should change this variable name?
+    :return: A list of strings which are to become the new chunks.
     """
     chunk_list = []  # container for chunks
     len_milestone = len(cutting_value)  # length of milestone term
-    cutting_value = cutting_value
+    cutting_value = cutting_value   # maybe we should delete this line?
```
chantisnake: use `# TODO: maybe we should delete this line?` so that we can search this in pycharm.
Thanks

PR #551, opened 2017-07-20
Feature/test cut by words

arialfie759:
- edited syntax of the comments of the cut_by_words function
- added type hinting to the function
- added pre-conditions to the function
- did unit testing on cut_by_words function
- created some error messages


test/unit_test/test_cut_word.py
```diff
@@ -0,0 +1,90 @@
+from lexos.helpers.error_messages import SEG_NON_POSITIVE_MESSAGE, \
+    OVERLAP_LARGE_MESSAGE, PROP_NEGATIVE_MESSAGE, OVERLAP_NEGATIVE_MESSAGE
+from lexos.processors.prepare.cutter import cut_by_words
+
+
+def test_cut_by_words():
+    assert cut_by_words(" ", 1, 0, .5) == [" "]
```
chantisnake: use more keyword parameter please

arialfie759: Would you like me to use keyword parameters for all of the parameters in the unit testing?

chantisnake: Yes, thanks

chantisnake: Actually this is not because it is in "unit testing", but because in current case "using keyword parameter is not awkward".
Remember what I said, "use keyword parameter unless it is too awkward"

PR #552, opened 2017-07-20
Bugfix/test cut line

Shi26:


lexos/helpers/error_messages.py
```diff
@@ -0,0 +1,5 @@
+NON_POS_NUM_MESSAGE = "the segment size should larger be positive"
```
chantisnake: Do not use `POS` here, since `POS` typically means `postion`

chantisnake:
nice

PR #561, opened 2017-07-25
move functions to lexos/interfaces

Landaluce:


chantisnake:
recreate pull request when your CI passes

PR #564, opened 2017-07-26
Refactor/interfaces

chantisnake:
- move all the functions in `lexos.py` into `interfaces`
- use flask's blueprint system to manage interfaces
- use `app.py` for the lexos app


scottkleinman:
Way ahead of schedule! Well done!

chantisnake:
I really cannot take any credit, this is such a amazing team. You should come here and see their energy.

We are getting things done faster than ever.

This is the best team i have ever worked with, I am really proud of them.

PR #568, opened 2017-07-27
X filename

mleblanc321:
* renamed the request header to be written as X-FILENAME (not X_FILENAME); this will also work when moved to be a public server


chantisnake:
I think this is a duplicate?
https://github.com/WheatonCS/Lexos/pull/559/files

chantisnake:
closing this for now

PR #570, opened 2017-07-27
Feature/change stats front

Weiqi97:
Description
* Applied new get_matrix method in generate_statistics function in managers/utility.py
  * Changed return type from dictionary to class instance
  * Changed naming style in interfaces/statistics_interface.py
  * Made corresponding changes in templates/statistics.html to have all data show up properly
* Made corresponding changes in processors/analyze/information.py
  * Applied more numpy method
  * Deleted redundant return function
* Fixed unit test and added more tests


Weiqi97:
Question
*  I realized that right now generate_statistics function is returning a list of class objects and a class object. So the type hinting right now should be wrong and I don't really know how to fix it.

chantisnake:
I will deal with that, thank you for fixing this function.

PR #572, opened 2017-07-28
feature/new_similarity

XinruLiu:
-convert all the array to numpy array in similarity.py
-add some changes to generate_similarities in utility.py 
to make it call function in similarity.py succesfully.
-change the unit test to test the similarity .py


lexos/managers/utility.py
```diff
@@ -1227,65 +1224,39 @@ def generate_similarities(file_manager: FileManager):
...
     # call similarity.py to generate the similarity list
-    docs_list_score, docs_list_name = similarity.similarity_maker(
-        count_matrix, comp_file_index, temp_labels)
+    docs_score, docs_name = similarity.similarity_maker(
+        final_matrix, comp_file_index, temp_labels)
 
     # error handle
-    if docs_list_score == 'Error':
-        return 'Error', docs_list_name
+    if docs_score == 'Error':
```
chantisnake: is this necessary?

test/unit_test/test_similarity.py
```diff
@@ -1,80 +1,74 @@
...
+        np.testing.assert_equal(
+            similarity_maker(count_matrix, comp_file_index, temp_labels),
+            (np.array([0.0, 0.0]), np.array(['catBobcat', 'test']))
+        )
 
     def test_with_two_dimension(self):
-        count_matrix = [['', 'cat', 'the'], ['file1', 0.0, 1.0],
-                        ['file2', 1.0, 2.0], ['file3', 2.0, 1.0]]
+        count_matrix = np.array([[0.0, 1.0], [1.0, 2.0],
+                                 [2.0, 1.0]], dtype=object)
```
chantisnake: dtype should not be object, `double` or `float` probably (I am not sure python has `double` type)

test/unit_test/test_similarity.py
```diff
@@ -1,80 +1,74 @@
...
         comp_file_index = 0
-        temp_labels = ['file2', 'file3']
-        assert similarity_maker(count_matrix, comp_file_index, temp_labels
-                                ) == ([0.1056, 0.5528], ['file2', 'file3'])
+        temp_labels = np.array(['file2', 'file3'])
+        np.testing.assert_equal(
```
chantisnake: for float probably should be assert close?

chantisnake: https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_allclose.html

PR #573, opened 2017-07-28
npmtest

scottkleinman:
This pull requests restores external javascript libraries inside a `static/node_modules`. HTML templates are updated to call these libraries from that folder. A procedure is implemented for installing libraries using `npm`. For a few javascripts and fonts it was not possible to use `npm`. These are placed in `static/js` and `static/fonts` respectively.


Shi26:
I am Cheng, Shi is using my laptop, therefore I took hers.

first I think `package-lock.json` should be ignored.
second I am not that comfortable that so many `TTF` file are in our repo.

Can I create a new branch and redo all of those? Thanks.

chantisnake:
as for lato font, I found this amazing repo, which will be able to install via `npm` or `brower`:

https://github.com/betsol/lato-font

Do you think this is good enough for your usage?

scottkleinman:
Let me make some adjustments and updated the branch. You may not need to make a new one.

scottkleinman:
I have now removed the Lato font since it is not currently being used. We can leave local implementation for a later fix.

chantisnake:
No, the more the commit, the larger the `.git` folder, therefore will cause a long clone.
**especially the change of binary files.**

merging your branch brings the commits involving binary file to the master, that will even enlarge current clone time (which is not fast right now, given the size of our project)

scottkleinman:
If I understand you correctly, the solution is to download a zip archive of the `npmtest` branch (which currently only has the Junicode binaries), create a new branch and unzip the old one into it, and then make it a single commit. Feel free to do that, if that's what's necessary.

chantisnake:
Yest, that is what needs to be done.

scottkleinman:
Are you able to do that quickly? Right now the menus are not fully functional in master because the necessary javascript is missing.

chantisnake:
Can we talk on slack? I have more question that I was meaning to ask you.

And if we talk about them here, we are spamming everyone else.

scottkleinman:
I'm ready to direct message you, but you're listed as "away" on Slack. Just send me a DM to start a conversation.

Sorry about the "spam", everybody.

chantisnake:
close this for now

PR #577, opened 2017-07-31
General functions documentation

Landaluce:
fixed documentation, deleted repeated or unused functions, added type hinting and unit test


chantisnake:
Unit test and style test both failed.

test/unit_test/test_general_functions.py
```diff
@@ -0,0 +1,78 @@
+from lexos.helpers.general_functions import get_encoding, make_preview_from, \
+    generate_d3_object, merge_list, load_stastic, matrix_to_dict, \
+    dict_to_matrix, html_escape, apply_function_exclude_tags, decode_bytes
+import unittest
+
+
+class TestGeneralFunctions(unittest.TestCase):
```
chantisnake: we actually do not use `unittest` framework of python, therefore inheriting `TestCase` will be unnecessary. Do you have any reason we should switch to `unittest` framework?

test/unit_test/test_general_functions.py
```diff
@@ -0,0 +1,78 @@
...
+        assert matrix_to_dict([['', 'a', 'b', 'c', 'd'], [0, 1, 2, 3, 4]]) == \
+            [{'a': 1, 'b': 2, 'c': 3, 'd': 4}]
+
+    def test_dict_to_matrix(self):
+        assert dict_to_matrix(
+            [{'a': 1, 'b': 2, 'c': 3, 'd': 4}]) == (
+            [['', 'a', 'b', 'c', 'd'], [0, 1, 2, 3, 4]],
+            ['a', 'b', 'c', 'd'])
+
+    def test_html_escape(self):
```
chantisnake: give it more test cases, put all the test function in one class

chantisnake: Actually the previous couple of functions like `test_dict_to_matrix` needs more test case too, but we are removing that function pretty soon, therefore you don't need to worry about them.

test/unit_test/test_general_functions.py
```diff
@@ -0,0 +1,78 @@
...
+    def test_apply_function_exclude_tags(self):
+        input_str = "<tag>asdf</tag>"
+
+        def dummy_function(input_string):
+            return input_string + input_string
+
+        assert apply_function_exclude_tags(
+            input_str, [dummy_function]) == "<tag>asdfasdf</tag>"
+
+    def test_decode_bytes(self):
```
chantisnake: this test case is not good, right?

test/unit_test/test_general_functions.py
```diff
@@ -0,0 +1,78 @@
...
+            ['a', 'b', 'c', 'd'])
+
+    def test_html_escape(self):
+        assert html_escape("&") == "&amp;"
+        assert html_escape('"') == "&quot;"
+        assert html_escape("'") == "&apos;"
+        assert html_escape(">") == "&gt;"
+        assert html_escape("<") == "&lt;"
+
+    def test_apply_function_exclude_tags(self):
```
chantisnake: please make more test case. Thank you.

and make sure the test case covers more extreme cases

test/unit_test/test_general_functions.py
Landaluce: which functions are going to be removed?

chantisnake: Those functions above, don't worry about them, just add more test case for these 3 functions (`test_html_escape`, `test_apply_function_exclude_tags`, `test_decode_bytes`).

Maybe add a class for each of them, and group the test cases using functions.

If you can read other people's test case that would be helpful to you.

test/unit_test/test_general_functions.py
```diff
@@ -0,0 +1,122 @@
...
+                                                                    'ASDF' \
+                                                                    '</tag>'
+
+    def test_multiple_functions(self):
+        assert apply_function_exclude_tags(
+            '<tag>asdf</tag>', [str.upper, str.lower,
+                                self.dummy_function]) == '<tag>asdfasdf</tag>'
+
+
+class TestDecodeBytes:
```
chantisnake: This function is well done.
But there are more encoding than this.

Here introduces couple of common encoding on the web: https://stackoverflow.com/questions/8509339/what-is-the-most-common-encoding-of-each-language

Test the first 3:
UTF-8    (89.2%)
ISO-8859-1   (5.0%)
Windows-1251 (1.6%)

and also `utf-8 with bom` and `utf-16 with bom`

test/unit_test/test_general_functions.py
```diff
@@ -0,0 +1,122 @@
...
+        assert html_escape("<html lang='en'></html>") == '&lt;html lang=&apo' \
+                                                         's;en&apos;&gt;&lt;' \
+                                                         '/html&gt;'
+        assert html_escape('<html lang="en"></html>') == '&lt;html lang=&quo' \
+                                                         't;en&quot;&gt;&lt;' \
+                                                         '/html&gt;'
+
+
+class TestApplyFunctionExcludeTags:
+    input_str = "<tag>asdf</tag>"
```
chantisnake: make another one with longer `input_str` please.

test/unit_test/test_general_functions.py
```diff
@@ -0,0 +1,122 @@
...
+            [{'a': 1, 'b': 2, 'c': 3, 'd': 4}]
+
+    def test_dict_to_matrix(self):
+        assert dict_to_matrix(
+            [{'a': 1, 'b': 2, 'c': 3, 'd': 4}]) == (
+                [['', 'a', 'b', 'c', 'd'], [0, 1, 2, 3, 4]],
+                ['a', 'b', 'c', 'd'])
+
+
+class TestHtmlEscape:
```
chantisnake: try 
- empty string 
- long string without any one of these characters.

lexos/helpers/general_functions.py
```diff
@@ -219,4 +219,4 @@ def decode_bytes(raw_bytes: bytes) -> str:
         # Grab the file contents, which were encoded/decoded automatically
         # into python's format
         decoded_string = raw_bytes.decode(encoding_type)
-    return decoded_string
+    return decoded_string.replace("\x00", "")
```
chantisnake: can you point out any source supporting this?

lexos/helpers/general_functions.py
```diff
@@ -140,20 +97,19 @@ def merge_list(word_lists):
     return merged_list
 
 
-def load_stastic(file):
-    """
-    this method takes an ALREADY SCRUBBED chunk of file(string), and convert
-    that into a WordLists
-    (see :return for this function or see the document for 'test' function)
+def load_stastic(input_string: str) -> dict:
+    """this method takes an ALREADY SCRUBBED chunk of file(string), and
```
chantisnake: cannot put change of line in the one line summary.

You can put multiline documentation after the empty line.

Please re-read the standard for documentation. Or reading other people's documentation can also be helpful

test/unit_test/test_general_functions.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def test_one_function(self):
+        assert apply_function_exclude_tags(
+            self.input_str, [self.dummy_function]) == '<tag>asdfasdf</tag>'
+        assert apply_function_exclude_tags(
+            self.input_str, [str.upper]) == '<tag>ASDF</tag>'
+
+    def test_two_functions(self):
+        assert apply_function_exclude_tags(
+            '<tag>asdf</tag>', [str.upper, self.dummy_function]) == '<tag>' \
```
chantisnake: why do you make a `input_str` and then do not reference it?

PR #589, opened 2017-08-02
Feature/eslint

chantisnake:
use eslint to check for js style errors:

```console
cd path/to/Lexos/lexos/statics
npm install
./node_module/.bin/eslint .
```

This will eventually be in CI, after we fixes all our js style errors

### DIFF
--- a/lexos/static/.eslintrc.json
+++ b/lexos/static/.eslintrc.json
@@ -1,0 +1,9 @@
+{
+  "extends": "standard",
+  "env": {
+    "browser": true,
+    "commonjs": true,
+    "es6": true,
+    "jquery": true
+  }
+}
--- a/lexos/static/.eslintrc.json
+++ b/lexos/static/.eslintrc.json
@@ -1,5 +1,10 @@
 {
   "extends": "standard",
+
+  "rules":{
+    "space-before-function-paren": "error"
+  },
+
   "env": {
     "browser": true,
     "commonjs": true,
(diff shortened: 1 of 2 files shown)


scottkleinman:
Why does the config file extend standard instead of eslint-recommended? What are the differences?

chantisnake:
unlike `jslint`, `eslint` actually do not have a recommended setting. (see http://eslint.org/docs/user-guide/getting-started)

Currently the most popular js standard are:
`standard` which is enforced by open source community. (this one is available is pycharm)
`google`, obviously enforced by google. (this one is available is pycharm)
`airbnb`, this one is actually not that popular, but is also adopted by eslint.

I personally prefer `standard` style, it is very clean, concise and functional.

What do you think?

scottkleinman:
I am annoyed by Standard's insistence on `function (response)`, instead of `function(response)`, which is pretty common in jQuery. But I can live with it. Otherwise, I'm fine with Standard's style. Eslint does have a minimal "recommended" set of rules which produced pretty good results in my editor. But I'm sure these rules don't enforce things like conventions for variable names. So we can stick with Standard for now.

chantisnake:
hum, that is hard to look...

I will see if there is anyway to fix that.

chantisnake:
I think the latest push fixes that

chantisnake:
but I do think anonymous function should have a space between "function" and the parenthesis

chantisnake:
also in the future we will probably use (arrow function)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions] instead of "anonymous function"

PR #594, opened 2017-08-02
Bugfix/interface documentation

Landaluce:


lexos/interfaces/base_interface.py
```diff
@@ -14,12 +13,13 @@
...
 
 
-def detect_active_docs():
-    """
-    This function (which should probably be moved to file_manager.py) detects
-    the number of active documents and can be called at the beginning of each
-    tool.
+def detect_active_docs() -> int:
+    """detects the number of active documents. This function can be called at
+    the beginning of each tool.
```
chantisnake: You cannot have a change of line in the one line summary.

Please re-read the documentation standards

chantisnake:
@scottkleinman, you seems to be doing well with git. I am very happy about that.

There is one thing, this branch still has some non-meaningful comment, like:
```
:returns: a response object (often a render_template call) to flask and 
eventually to the browser. 
```

Therefore, it will be very more nice of you to review the branch more carefully before merging.
Nevertheless merging into `master` is kind of a big thing. We want to try our best to make sure that the code change is at a reasonable quality.

Thank you.
I will try to unmerge this branch. See if I can do that.

PR #597, opened 2017-08-03
Fixed session handling and UI bugs for tag handling

scottkleinman:


scottkleinman:
I can't get past this PEP8 violation, which is unrelated to the changes I made. Even when I copy the raw code from master, I still get the error. Help?

scottkleinman:
OK, running my own PEP8 script revealed the problem. But the code could use another pair of eyes before we merge.

chantisnake:
If you want to reproduce the error on CI, try the following command:
```bash
cd path/to/Lexos
pip install -r ./dev_requirements.txt
flake8 --ignore F841,E123 --count .
```

lexos/templates/scrub.html
```diff
@@ -34,6 +34,7 @@
                 }
                                 });
            tagDict=JSON.stringify(tagDict);
+           console.log(tagDict);
```
chantisnake: Question, is it actually good that we put so many `console.log` on the frontend.

scottkleinman: I don't know that it actually has a noticeable effect on performance, but it's technically debugging code and so should probably be commented out when not in use. And we should probably have a pre-release step of removing stray `console.log` statements, just as we do with `print` commands in Python.  But I think the time to adopt this practice is when we begin our first Javascript cleanup.

chantisnake: I mean, in python we have stopped using `print` statement, and just use debugger to debug.

I think there are great debugger for the front-end too, what editor and plugin are you using?

scottkleinman: Mostly Sublime and occasionally Visual Studio Code. I just installed he eslint plugins, but I haven't explored front-end debuggers.

chantisnake: I think vscode has chrome debuggers, I am not sure it is great. I am pretty positive sublime text will also have chorme debugger and firefox debugger.

chantisnake: one of the plugin that will make it easier for you to pass the CI is `editorconfig`. we have an editor config file in the root of our project, this takes care of final new line and trailing white spaces as well as indentation

scottkleinman: I will investigate browser debuggers. `editorconfig` is already installed in vs code, but I just added it to Sublime.

chantisnake: personally I believe, (I am probably biased, I never loved sublime-text), vscode provides a better debugging experience

PR #603, opened 2017-08-04
Feature/flask error handler

chantisnake:
- implemented the backend for flask error handling
- a new 404 and 500 page
- fix the indentation in all the js file, because they are making me crazy
- add the `babel` in package the json (this helps us in the future to compile es6 to es5) (probably do not belong in this branch, but It will not cause conflict with anything)

- ajaxes are **not** implemented with `LexosException`, because currently ajaxes are too messy, I will change them later


scottkleinman:
I have not problem with the style fixes in the Javascript, but for `jquery.dataTables.areaselect.js` and `natural.js` are technically other people's code which we were simply unable to install through `npm`. So we may not want to change those files. It may even be worth it to segregate them in a `vendors` folder.

chantisnake:
the thing is, they are not in the `node_modules` folder, therefore most of the editor and checker will automatically thinks that is our code, and probably will not be happy about it.

And I only changes the indentation and formatting (vscode auto format all these documents), this should not break their code.

PR #612, opened 2017-08-07
Lato import

scottkleinman:
This branch implements a local import of the Lato font, which is much faster and more reliable than doing it from Google Fonts. It's worth importing Lato this way, even if not in `LOCAL_MODE`. Note that the import clones exactly the configuration we had from Google Fonts, which Google Fonts lists as having a "moderate" download time. We could possibly shave Bold, Italic, Black variants we are not using to speed loading times even more.


chantisnake:
I thought we have already settled on the solution to use npm to install lato font?

chantisnake:
I personally would love to minimize the number of binary file on our repo.

scottkleinman:
I think that I may have blocked this out because the issue was purely theoretical since we couldn't actually import the fonts locally. I have figured out how to do that now. Let's hold off on merging this branch, and I'll experiment with the npm version. It needs to be tested because it is a later version of the font. I'll test and then update this pull request.

chantisnake:
thank you

scottkleinman:
OK, the latest changes pull the Lato font from npm, but there are some caveats:

1. It is version 2 of the Lato font, which appears to be slightly smaller. It is difficult, and possibly impossible to get the font to appear exactly as it does using the version on Google Fonts, which is version 1. To see the difference, run this branch in a different tab from the one on the server and toggle between them. We'll need to run this by Mark to see if he finds it acceptable, since it affects the banner.

2. The npm version does nto appear to have separate Latin and extended Latin versions of the font. It is possible that this will affect the display of some non-ASCII characters, but I'm not sure how to test this.

chantisnake:
you probably need to remove `package-lock.json`. Thank you.

scottkleinman:
Forgot about that. It's done.

chantisnake:
Hum, I still see it in the file changes list... https://github.com/WheatonCS/Lexos/pull/612/files

scottkleinman:
Sorry about that. Looks like I forgot to push the commit. A fire alarm went off at 7:35 this morning (4:35 my time), so I am not at my best...

chantisnake:
That is totally fine

scottkleinman:
I'm not going to merge until Mark takes a look to make sure that he is happy with the font.

chantisnake:
okay

chantisnake:
It is generally not that big a deal, but the best practice is probably waiting for all 4 CI to pass and then merge.

Appveyor is always a pain in the ass, but we do need our app to work on windows.

scottkleinman:
Sorry about that. I ran the changes past Mark, and he said, "Do you want to merge now", and I did so impulsively without looking. But my practice has been to wait for Appveyor, as annoying as it is...

chantisnake:
sure, I understand

PR #640, opened 2017-08-17
Feature/implement webassets

chantisnake:


chantisnake:
- by "lexos css", do you mean `css/style.css`? I think that is in the bundle: see [here]( https://github.com/WheatonCS/Lexos/blob/5a55a06f49d119a5e8dc6ba4d3634a32aa248995/lexos/static/webasset_bundle_css.yml#L3)

- I will investigate more on loading unbundled assets, but the intuitive way is to simply add the link to the html

scottkleinman:
I was referring to `css/style.css` (and there might be some others marked with a Lexos CSS comment in the template file). I missed that you had put them in the bundle, but I see that now. Bundling these files may not be a good idea, depending on whether we an load them unbundled for development purposes.

On that note, `assets_env.debug = True` may be the answer. It's under "Configuration" in the [Flask-Assets docs](https://flask-assets.readthedocs.io/en/latest/).

I'm not sure what the implications of the file sizes are.

chantisnake:
- `style.css` is not rendered because I put at first, therefore I guess it is overloaded by other css file. it is fixed now

- `assets_env.debug = True` prevents webassets to generate the hash of the build file. 
  - if debug is true, we will have file `all.js?fd1b674f`
  - if debug is false, we will have file `all.js`

- the implecation of the file size is: even if we minimize the js and css file by bundle, we still have a pretty large file to send. Therefore webassets (more precisely `cssmin` and `jsmin`) is not that efficient.

chantisnake:
Uh, I finally understand what you mean by debug mode. You need the js and css code to be "un-filtered" to debug the js and css code.

Then `assets_env.debug = True` is the solution

scottkleinman:
Are most of the `file not found` errors in the CSS? This can be a problem because identifying the proper relative paths can be difficult with Blueprints. Maybe webassets is magnifying the problem?

Regarding the efficiency of the minimisation, even if the file size is large, is the user experience better with a single http request? That's the theory, anyway. There is a moderate improvement in terms of the simplicity of the code that goes in the template, but you pay a price with greater complexity on the back end. So, given what you have discovered so far, do you think it is worth pursuing this approach?

chantisnake:
- many of them are js, most of them are simple, but they are pretty spread out on each page.

- I don't know what is the standard in the industry, but leaving the CDN link there seems to be too complicated for me.

- I currently don't have much faith in `webassets` because of [this issue](https://github.com/miracle2k/webassets/issues/291). This will greatly hinder our development in the front end, and it haven't been fixed in 3 years...

Issue #639, opened 2017-08-16
Improved Gutenberg detection


EmmaGrace:
Right now, we are attempting to detect Project Gutenberg boilerplate on uploaded files, and place a very intrusive dialogue box on the scrubber page if we do so. The boilerplate detection is handled by the LexosFile class, and the derived boolean is then passed to scrub(), which will then automatically attempt to remove the boilerplate. However, LexosFile's check_for_gutenberg() method is a bit crude.

The only factor that determines whether that giant pop-up occurs is if the uploaded text contains the string "Project Gutenberg" exactly as typed, case sensitive. This means a text discussing Project Gutenberg will trigger that pop-up, and any Gutenberg boilerplate that has been scrubbed to lowercase in a different program will not. We easily simply make our check non-case sensitive, but a researcher who works regularly with Gutenberg texts could probably clue us in to a more significant pattern to look for.

Landaluce:
This can be fixed by searching for "\*\*\* START OF THIS PROJECT GUTENBERG.*?\*\*\*" which is what scrubber is using to find gutenberg's front boiler plate

chantisnake:
A PR is merged. I am gonna assume this is fixed.

PR #642, opened 2017-08-17
Feature/improve gutemberg detection

Landaluce:


chantisnake:
just to mark this is the pr for issue: https://github.com/WheatonCS/Lexos/issues/639

lexos/managers/lexos_file.py
```diff
@@ -128,7 +128,8 @@ def check_for_gutenberg(file_contents: str) -> bool:
         :return: a boolean representing if file is from Project Gutenberg.
         """
 
-        if re.search('Project Gutenberg', file_contents):
+        if re.search("\*\*\* START OF THIS PROJECT GUTENBERG.*?\*\*\*",
+                     file_contents):
```
mleblanc321: should the message be stored in a constant?

mleblanc321: kind of a hack, not sure how prevalent this "START OF ..." message is @gutenberg ?

chantisnake: I agree this is a hack. We should use XML.

scottkleinman: @chantisnake, I'm not sure how we would use XML for this. Can you explain?

chantisnake: Hum, I am guessing that they have a specific XML tag at a specific place?

I can be wrong.

scottkleinman: Sadly, they do not. The text is either plain text or HTML, but the HTML is not even that helpful.

chantisnake: too bad, I need more investigation to give a definite answer to this problem

Landaluce: they don't have any tags, and this is the only part of the Gutenberg boilerplates that is the same for all the Gutenberg files

chantisnake:
I think this is an okay temporary solution. I think still think this is Hacky.

I personally think we need at least add a box on front-end to toggle whether to remove the boiler-plate.

scottkleinman:
The pop-up warning _is_ very obtrusive. Perhaps something like this would provide a better user experience:

image


If anyone wants to do this, I got the image above by injecting the following code in the browser console right before the row containing the `mainscruboptions` div:

```html
<div class="row">
    <div class="col-md-10">
    <label style="color:red">Your documents appear to contain Project Gutenberg boilerplate.</label>
    <br>
    <label id="tmp" for="tmpbox" class="icon-checkbox checked">    
      <input type="checkbox" name="tmpbox" id="tmpnbox" checked="">Attempt to Remove Boilerplate<i class="fa fa-question-circle lexos-tooltip-trigger" data-toggle="tooltip" data-html="true" data-placement="right" data-container="body" title="Project Gutenberg boilerplate is annoying." style="padding-left:4px;font-size:18px;"></i>
     </label>
    </div>
</div>
```

PR #651, opened 2017-09-11
Feature/matrix model

chantisnake:


chantisnake:
👍

lexos/managers/file_manager.py
```diff
@@ -1290,8 +1297,8 @@ def delete_all_file(self):
 
     # Experimental for Tokenizer
     @staticmethod
-    def get_matrix_options_from_ajax() -> Tuple[int, bool, bool, bool, str,
-                                                bool, bool, bool, bool, bool]:
+    def get_matrix_options_from_ajax_deprec() -> \
+            Tuple[int, bool, bool, bool, str, bool, bool, bool, bool, bool]:
```
EmmaGrace: Use parens instead of Tuple

chantisnake: nice catch, I am going to fix that

PR #652, opened 2017-09-13
feature/clean_up_kmeans

Weiqi97:
**Description**
* Minor bug fixed for PCA method at the front end.
* Cleaned up Kmeans.py
  * Changed documentations.
  * Removed unnecessary helper functions.
  * Cleaned up code for Kmeans in utility and applied panda data frame.
  * Changed Kmeans.py to two class structures and instead of returning a bunch of stuff, return the class instance and throw it to the front.
  * Made change corresponding to the structure change in utility.py and in the interface.
  * Added unit test.


Weiqi97:
I'll add more tests in unit test for special cases. Should be done by tomorrow.

chantisnake:
nice job

chantisnake:
There are some deprecations added in #651, therefore it is possible that you need to change some of your function names after we merge #651 to `master`. 

One example would be `get_matrix` function is deprecated (again) in favor of matrix model.

lexos/helpers/constants.py
```diff
@@ -161,6 +161,11 @@
     'tolerance',
     'KMeans_metric',
     'viz')
+# for kmeans analyzing
+N_INIT = 300
+MAX_ITER = 300
+TOLERANCE = 1e-4
```
mleblanc321: can we have some documentation on these three values?  e.g.,  TOLERANCE (of what?)

PR #655, opened 2017-09-25
Basic prototype for Dendro model

chantisnake:
- some fixes in `matrix_model`
- new receivers structure
- a prototype for dendrogram model
  - manage to create the same topology in DAZ
  - much cleaner backend and frontend code
  - completely remove Silouhiette Score
  - does not look pretty, needs further work on this
  - download is not implemented
- Still no test is implemented... I promise, next week, I promise

### DIFF
--- a/lexos/application.py
+++ b/lexos/application.py
@@ -2,7 +2,6 @@ import json
 import os
 import re
 import time
-import traceback
 
 from flask import Flask, request, render_template
 from jinja2 import evalcontextfilter


scottkleinman:
What is the justification behind removing the silhouette score?

chantisnake:
@mleblanc321 thinks that silhouette score is not a good enough measure.

mleblanc321:
we basically don't use it, we know barely how it works, cheng is dubious about the code ... lots of reasons to just remove it from the tool

scottkleinman:
Have you read [http://scalar.usc.edu/works/lexos/silhouette-scores](http://scalar.usc.edu/works/lexos/silhouette-scores)?

Silhouette scores are currently the only method we have of evaluating the robustness of clusters. I wonder if we could leave them with a better explanation of their limitations.

chantisnake:
@mleblanc321 thinks that `boot-strap consensus tree`(@Weiqi97 is currently implementing) is a better measure in terms of the dendrogram.

I agree with Mark, since clustering dendrogram is somewhat hard.

For k-means, I have no problem keeping Silhouette Score.

PR #660, opened 2017-10-30
Feature/content analysis

Landaluce:
Integrate dictionary-base content analysis tool


lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,327 @@
...
+
+        dictionaries: List of Dictionary objects
+        corpus: List of File objects
+        counters: a 2D array with count of every dictionary for evey file in
+        the corpus
+        formulas: List of string that represent a formula for each file
+        scores: List of formula/total word count of each file
+        averages: Lis of averages count of each dictionary
+        """
+        self.dictionaries = []
```
chantisnake: make them private if you can

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,327 @@
...
+        self.dictionaries.append(Dictionary(content=new_list,
+                                            file_name=file_name,
+                                            label=label))
+
+    def delete_dictionary(self, label: str):
+        """deletes a dictionary
+
+        :param label: label of dictionary to delete
+        """
+        for i in range(len(self.dictionaries)):
```
chantisnake:
```python
new_dicts = [dictionary for dictionary in self.dictionaries if dictionary.label != del_label]
self.dictionaries = new_dicts
```

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,327 @@
...
+        for dictionary in self.dictionaries:
+            if dictionary.label == label:
+                dictionary.active = not dictionary.active
+
+    def get_active_dicts(self) -> list:
+        """
+
+        :return: a list containing all active dictionaries
+        """
+        active_dicts = []
```
chantisnake:
```python
return [dictionary for dictionary in self.dictionaries if dictionary.active]
```

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,327 @@
...
+        for dictionary in self.dictionaries:
+            if dictionary.active:
+                active_dicts.append(dictionary)
+        return active_dicts
+
+    def detect_active_dicts(self) -> int:
+        num_active_dicts = 0
```
chantisnake:
```python
return len(self.get_active_dicts)
```

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,327 @@
...
+    def toggle_dictionary(self, label: str):
+        """Activates and Deactivates a dictionary
+
+        :param label: filename of dictionary to toggle
+        """
+        for dictionary in self.dictionaries:
+            if dictionary.label == label:
+                dictionary.active = not dictionary.active
+
+    def get_active_dicts(self) -> list:
```
chantisnake: ` -> List[Dictionary]` be more specific on types

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,327 @@
...
+
+    def detect_active_dicts(self) -> int:
+        num_active_dicts = 0
+        for dictionary in self.dictionaries:
+            if dictionary.active:
+                num_active_dicts += 1
+        return num_active_dicts
+
+    def count_words(self):
+        """counts all dictionaries for all active files in the corpus
```
chantisnake: not the way to write one-line comment

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,327 @@
...
+        result = 0
+        for i in range(len(self.corpus)):
+            new_formula = formula
+            for j in range(len(active_dicts)):
+                new_formula = new_formula.replace(
+                    "[" + active_dicts[j].label + "]",
+                    str(self.counters[i][j]))
+            new_formula = new_formula.replace("()", "")
+            try:
+                result = eval(new_formula)
```
chantisnake: check whether the formula is a valid mathematical expression.

lexos/static/js/scripts_content_analysis.js
```diff
@@ -0,0 +1,165 @@
+/**
+ * Created by alvaro on 9/23/17.
+ */
```
chantisnake: add document to your javascript.

you just type `/**<RET>` pycharm will generate the document template

lexos/models/content_analysis_model.py
```diff
@@ -2,9 +2,10 @@
 
 
 # do not delete! used in generate_scores() by eval()
-# noqa F401
+
 # noinspection PyUnresolvedReferences
-from math import sqrt, sin, cos, tan, log
+# noqa F401
+from math import sqrt, sin, cos, tan, log  # NOQA
```
chantisnake: change the `#NOQA` to `noqa F401`

lexos/models/content_analysis_model.py
```diff
@@ -2,9 +2,10 @@
 
 
 # do not delete! used in generate_scores() by eval()
-# noqa F401
+
 # noinspection PyUnresolvedReferences
-from math import sqrt, sin, cos, tan, log
+# noqa F401
```
chantisnake: remove this line

lexos/views/content_analysis_view.py
```diff
@@ -0,0 +1,209 @@
...
+                               dictionary_labels=dictionary_labels,
+                               active_dictionaries=active_dictionaries,
+                               toggle_all=session['toggle_all'])
+    else:
+        formula = request.json['calc_input']
+        if len(formula) == 0:
+            session['formula'] = "0"
+        else:
+            formula = formula.replace("√", "sqrt").replace("^", "**")
+            session['formula'] = formula
```
chantisnake: don't use session, use requests

lexos/views/content_analysis_view.py
```diff
@@ -0,0 +1,209 @@
...
+            error_msg = "Formula errors:<br>"
```
chantisnake: if you want to send an error to the frontend, raise an "LexosException", see this function: https://github.com/WheatonCS/Lexos/blob/master/lexos/application.py#L111

lexos/views/content_analysis_view.py
```diff
@@ -0,0 +1,209 @@
...
+from lexos.models.content_analysis_model import ContentAnalysisModel
+from lexos.views.base_view import detect_active_docs
+
+# this is a flask blue print
+# it helps us to manage groups of views
+# see here for more detail:
+# http://exploreflask.com/en/latest/blueprints.html
+# http://flask.pocoo.org/docs/0.12/blueprints/
+content_analysis_blueprint = Blueprint('content_analysis', __name__)
+analysis = None
```
chantisnake: no global variable please.

lexos/views/content_analysis_view.py
```diff
@@ -0,0 +1,209 @@
...
+        for file in active_files:
+            analysis.add_corpus(file_name=file.name,
+                                label=file.label,
+                                content=file.load_contents())
+    if request.method == 'GET':
+        dictionary_labels = []
+        active_dictionaries = []
+        session['toggle_all'] = False
+        if len(analysis.dictionaries):
+            session['toggle_all'] = True
```
chantisnake: try not to access session outside of session manager

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,350 @@
...
+        self._averages.append(sums_avg)
+        self._averages.append(total_word_counts_avg)
+        self._averages.append(scores_avg)
+
+    def join_active_dicts(self) -> list:
+        active_dicts = self.get_active_dicts()
+        dictionaries = [Phrase(content=phrase, label=dictionary.label)
+                        for dictionary in active_dicts
+                        for phrase in dictionary.content]
+        dictionaries.sort(key=lambda x: len(x.content.split()), reverse=True)
```
EmmaGrace: this probably needs a comment to explain

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,350 @@
...
+            if len(self._counters) != 0:
+                self._averages.append(round(
+                    float(cat_count) / len(self._counters), 1))
+            else:
+                self._averages.append(0)
+        self._averages.append(sums_avg)
+        self._averages.append(total_word_counts_avg)
+        self._averages.append(scores_avg)
+
+    def join_active_dicts(self) -> list:
```
EmmaGrace: missing docstring

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,350 @@
...
+        active_dicts = self.get_active_dicts()
+        dictionaries = [Phrase(content=phrase, label=dictionary.label)
+                        for dictionary in active_dicts
+                        for phrase in dictionary.content]
+        dictionaries.sort(key=lambda x: len(x.content.split()), reverse=True)
+        return dictionaries
+
+    def to_html(self) -> str:
```
EmmaGrace: missing docstring

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,350 @@
...
+        :return: a data frame containing all values stored in this class
+        members
+        """
+        columns = ['Document Name'] + [dictionary.label for dictionary in
+                                       self.get_active_dicts()] + \
+                  ['formula', 'total word count', 'score']
+        indices = [file.label for file in self._corpus] + ['averages']
+        df = pd.DataFrame(columns=range(len(columns)),
+                          index=range(len(indices)))
+        for i in range(len(self._corpus)):
```
EmmaGrace: this section needs comments, better var names, and enumerate

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,350 @@
...
+                df.xs(i)[j + 1] = self._counters[i][j]
+            df.xs(i)[j + 2] = self._formulas[i]
+            df.xs(i)[j + 3] = self._corpus[i].total_word_counts
+            df.xs(i)[j + 4] = self._scores[i]
+        for j in range(len(self._averages)):
+            df.xs(i + 1)[j] = self._averages[j]
+        df.columns = columns
+        return df
+
+    def is_secure(self, formula: str):
```
EmmaGrace: needs docstring

chantisnake:
Hi, alvaro. 

I saw you working hard on this problem. I understand creating a new functionality is really hard. You need to actually design the functionality really carefully. 

I think you have read our paper. which explains why I don't want to mutate variable or to write loops. Find another way to approach the problem please. 

Don't afraid to call functions, you don't need to access the class member every time, just call a function to generate all the data.

For example the "add_corpus` function (wouldn't `add_file` be more appropriate?) why can't you just directly get them from the `file_manager` every time you use them? Just like what we do in our matrix model: https://github.com/WheatonCS/Lexos/blob/bugfix/improve_dendro_and_matrix/lexos/models/matrix_model.py#L37

Same thing for the count. you cache all your counts in your members. what if people mutate it in the future? why not simply return your counts from the `count` function?


Next time you working on large software. Remember, how fast you code is not the way to evaluate your performance. Think about your implementation first. 
Don't rush it. Even Google's engineer on average write 100 lines of code per day.

Elegant code is much more important than more code. 
Inelegant code sometimes is worse than no code.

Hope this post helps you in some way. 

Best,
Cheng

chantisnake:
If you actually think about your code:
- you need people to add documents before they call a count.
- you need people to count before they can access the count result.
- you also needs people to save the formula to evaluate the formula.

The people who want to use your code require a very comprehensive understanding of your code. they cannot simply look at the documentation for one function and then use that function. They need to know every single function (method) in your class to create a bug free program.

Is that intuitive? Is that a good code base to leave for the new developer to work with?

Not to mention all the returns you will get when you delete a dictionary. That is also really unintuitive.

Think about this for some time. Read through some other models. What do you think?

chantisnake:
I am not saying what you have here is not good. This is a good functionality and you made work. That is really impressive.

I just hope you understand why do we really need you to change your code base.

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,465 @@
...
+            else:
+                self._averages.append(0)
+        self._averages.append(sums_avg)
+        self._averages.append(total_word_counts_avg)
+        self._averages.append(scores_avg)
+
+    def join_active_dicts(self) -> list:
+        """
+
+        :return: list of phrases contained in the active dictionaries
```
chantisnake: This a single line documentation: you need to follow one line documentation syntax: https://github.com/WheatonCS/Lexos/wiki/Python-Coding-Style-Guide#comment-syntax

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,465 @@
...
+        dictionaries = [Phrase(content=phrase, dict_label=dictionary.label)
+                        for dictionary in active_dicts
+                        for phrase in dictionary.content]
+        dictionaries.sort(key=lambda x: len(x.content.split()), reverse=True)
+        return dictionaries
+
+    def to_html(self) -> str:
+        """
+
+        :return: dataframe table in html
```
chantisnake: This is a single line documentation

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,465 @@
...
+        """
+        df = self.to_data_frame()
+        html = df.to_html(classes="table table-striped table-bordered",
+                          index=False)
+        return html
+
+    def to_data_frame(self) -> pd.DataFrame:
+        """
```
chantisnake: Please provide a summary to your documentation

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,465 @@
...
+                         " ", "+", "-", "*", "/", "sin", "cos", "tan", "log",
+                         "sqrt", "(", ")"]
+        for item in allowed_input:
+            formula = formula.replace(item, "")
+        if len(formula) == 0:
+            return True
+        return False
+
+    def save_formula(self):
+        """saves the formula from the front-end in the class member
```
chantisnake: summary can only have one line

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,465 @@
...
+            is_error = True
+        if "log()" in self._formula:
+            error_msg += "log takes exactly one argument (0 given)<br>"
+            is_error = True
+        if is_error:
+            return error_msg
+        return ""
+
+    def analyze(self) -> [str, str]:
+        """Calls all the functions needed to generate the final html table
```
chantisnake: summary can only have one line

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,465 @@
...
+        return self._scores
+
+    @property
+    def averages(self) -> list:
+        return self._averages
+
+    @property
+    def content_analysis_option(self) -> ContentAnalysisOption:
+        """
+        :return: front-end or test options
```
chantisnake: use one line commenting syntax

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,465 @@
...
+
+    @property
+    def name(self):
+        return self._name
+
+
+class Dictionary(Document):
+    def __init__(self, content: list, file_name: str, label: str,
+                 active: bool = True):
+        """
```
chantisnake: provide a summary to your documentation

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+    def generate_scores(self):
+        """calculate the formula and scores=formula/total_word_count for each
+        file in the corpus
+
+        """
+        self._scores = []
+        self._formulas = []
+        active_dicts = self.get_active_dicts()
+        result = 0
+        for i in range(len(self._corpus)):
```
EmmaGrace: Please enumerate instead

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+            new_formula = self._formula
+            for j in range(len(active_dicts)):
```
EmmaGrace: Enumerate, and you need to explain what this does with the formula

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+                new_formula = new_formula.replace(
+                    "[" + active_dicts[j].label + "]",
+                    str(self._counters[i][j]))
+            new_formula = new_formula.replace("()", "")
```
EmmaGrace: Why?

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+            try:
+                result = eval(new_formula)
+            except (ValueError, SyntaxError):
+                pass
```
EmmaGrace: It's bad to let errors pass silently. Does the user even know their formula did nothing?

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+            self._scores.append(round(
+                float(result) / self._corpus[i].total_word_counts, 3))
```
EmmaGrace: Why is there a 3 here?

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+            total_word_counts_sum += \
+                file.total_word_counts
+            formulas_sum += formula
+        if len(self.scores) != 0:
+            scores_avg = round(
+                (float(scores_sum) / len(self.scores)), 3)
+        else:
+            scores_avg = 0
+        if len(self._corpus) != 0:
+            total_word_counts_avg = round((float(total_word_counts_sum) /
```
EmmaGrace: Why 1?

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+                                           (len(self._corpus))), 1)
+        else:
+            total_word_counts_avg = 0
+        if len(self._formulas) != 0:
+            sums_avg = round((float(formulas_sum) / len(self._formulas)), 1)
+        else:
+            sums_avg = 0
+        self._averages.append("Averages")
```
EmmaGrace: You seem to be appending a string label to what I assume is a list of floats/ints. This is generally a bad practice, because it obfuscates the code. Is there a cleaner way you could accomplish your goal here?

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+        for x in range(len(active_dicts)):
```
EmmaGrace: Enumerate please

lexos/processors/analyze/content_analysis_model.py
EmmaGrace: After seeing this multiple times, it occurs to me that you are rounding to one decimal place. This is why you need to use keyword parameters, especially with messy function calls like this

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+                (float(scores_sum) / len(self.scores)), 3)
+        else:
+            scores_avg = 0
+        if len(self._corpus) != 0:
+            total_word_counts_avg = round((float(total_word_counts_sum) /
+                                           (len(self._corpus))), 1)
+        else:
+            total_word_counts_avg = 0
+        if len(self._formulas) != 0:
+            sums_avg = round((float(formulas_sum) / len(self._formulas)), 1)
```
EmmaGrace: keyword parameters

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+        else:
+            sums_avg = 0
+        self._averages.append("Averages")
+        for x in range(len(active_dicts)):
+            cat_count = sum([counter[x] for counter in self._counters])
+            if len(self._counters) != 0:
+                self._averages.append(round(
+                    float(cat_count) / len(self._counters), 1))
```
EmmaGrace: keyword parameters

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+            else:
+                self._averages.append(0)
+        self._averages.append(sums_avg)
+        self._averages.append(total_word_counts_avg)
+        self._averages.append(scores_avg)
```
EmmaGrace: You are appending to what was an empty list. I see no reason why you couldn't use list comprehension instead. Please do this wherever possible, instead of appending

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+        """
+        columns = ['Document Name'] + [dictionary.label for dictionary in
+                                       self.get_active_dicts()] + \
+                  ['formula', 'total word count', 'score']
+        indices = [file.label for file in self._corpus] + ['averages']
+        df = pd.DataFrame(columns=range(len(columns)),
+                          index=range(len(indices)))
+        # adds row to df with: file label, count for each dict,
+        # formula result, total word count, score
+        for i, (file, formula, score, counter) in enumerate(
```
EmmaGrace: This function is very hard to read. Some of this may be due to bad function names (xs), but please don't make it worse than it needs to be with your own single and double letter variable names!

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+            df.xs(i)[j + 3] = file.total_word_counts
+            df.xs(i)[j + 4] = score
+        # add a row to df with the average of each column
+        for j, (average) in enumerate(self._averages):
+            df.xs(i + 1)[j] = average
+        df.columns = columns
+        return df
+
+    def is_secure(self) -> bool:
+        """Checks if the formula is secure
```
EmmaGrace: Please define "secure" more explicitly. Looks like you mean something like "containing only names of uploaded dictionaries and approved math symbols"

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+            self._formula = "0"
+        else:
+            formula = formula.replace("√", "sqrt").replace("^", "**")
+            self._formula = formula
+        return self.check_formula()
+
+    def check_formula(self) -> str:
+        """Checks if there are any errors in the formula
+
+        :return: error message if there is an empty str or None if there is
```
EmmaGrace: If you were just checking for an empty string, this would be a very short function. Please describe what conditions actually trigger an error message

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+
+    @dictionaries.setter
+    def dictionaries(self, dictionaries: list):
+        self._dictionaries = dictionaries
+
+    @property
+    def corpus(self) -> list:
+        return self._corpus
+
+    @property
```
EmmaGrace: Most of these properties should probably be in a NamedTuple. If not, you need docstrings

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+    def test_option(self):
+        return self._test_options
+
+    @test_option.setter
+    def test_option(self, options):
+        self._test_options = options
+
+
+def count_phrases(dictionary, file):
```
EmmaGrace: Now I know that counting phrases is a tricky thing to do, because we discussed it at length in our meeting. You should really be explaining exactly how it works, with all the implications, somewhere in here

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+    """counts each phrase in the dictionary in the given file
+
+    :param dictionary: list of Phrase objects
+    :param file: a File object
+    :return: list of Phrase objects with their counts
+    """
+    for phrase in dictionary:
+        count = 0
+        if file.content.startswith(phrase.content + " "):
```
EmmaGrace: Please explain the cases you are handling here

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+                                 total_word_counts=total_word_counts))
+
+    def add_dictionary(self, file_name: str, label: str, content: str):
+        """Adds a dictionary
+
+        :param file_name: name of the file
+        :param label: label of the file
+        :param content: content of the file
+        """
+        new_list = str(content).split(", ")
```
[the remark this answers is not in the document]
Landaluce: what need to be explained here? A dictionary has a name, a label, and content (see class definition)

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+
+        :return: a list containing all active dictionaries
+        """
+        return [dictionary for dictionary in self.dictionaries
+                if dictionary.active]
+
+    def count(self):
+        """counts all dictionaries for all active files in the corpus"""
+        self._counters = []
+        dictionaries = self.join_active_dicts()
```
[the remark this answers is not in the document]
Landaluce: doesen't the docstring explain it?

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+        for file in deepcopy(self._corpus):
+            dictionaries = count_phrases(dictionaries, file)
+            self.get_dictionary_counts(dictionaries)
+
+    def get_dictionary_counts(self, dictionaries):
+        """gets the counts for each dictionary
```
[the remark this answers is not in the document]
Landaluce: gets the counts for each dictionary (see docstring)

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+
+        :param dictionaries: list of Phrase object
+        """
+        counter = []
+        active_dicts = self.get_active_dicts()
+        for dictionary in active_dicts:
+            count = 0
+            for phrase in dictionaries:
```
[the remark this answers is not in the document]
Landaluce: initially each Phrase is count, this loops add the count of each dictionary for each file

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,461 @@
...
+                if phrase.dict_label == dictionary.label:
+                    count += phrase.count
+            counter.append(count)
+            if len(counter) == len(active_dicts):
```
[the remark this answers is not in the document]
Landaluce: creates a list, where each member represents the cunts of all dictionaries for one file, so each index in the list represents a row of the table

lexos/processors/analyze/content_analysis_model.py
Landaluce: if the user inputs "()" as formula, is not an error, so I just remove those parenthesis

lexos/processors/analyze/content_analysis_model.py
Landaluce: to round to the first decimal

Landaluce: what's messy?

lexos/processors/analyze/content_analysis_model.py
Landaluce: to round to the first decimal

lexos/processors/analyze/content_analysis_model.py
Landaluce: if you see the result table, there is a row called Averages that contains the averages of each columns. Eventually, I have to append that string to the list of floats to be able to generate that row

lexos/processors/analyze/content_analysis_model.py
Landaluce: this would look like this: 
for i, active_dict in enumerate(active_dicts): 
but active_dict will never get used

lexos/processors/analyze/content_analysis_model.py
Landaluce: again, rounding

lexos/processors/analyze/content_analysis_model.py
Landaluce: I'm not sure how to do that, inside the loop there is a line that always gets executed and an if/else statement

lexos/processors/analyze/content_analysis_model.py
Landaluce: the "xs" is panda's code, not mine

lexos/processors/analyze/content_analysis_model.py
Landaluce: as the error messages say: it checks for mismatched parenthesis, and correct number of arguments for all allowed functions

lexos/processors/analyze/content_analysis_model.py
Landaluce: @chantisnake do we really need docstrings for getters and setters?

lexos/processors/analyze/content_analysis_model.py
Landaluce: what cases are you referring too?

lexos/processors/analyze/content_analysis_model.py
Landaluce: my bad, this try/except is not needed anymore

lexos/processors/analyze/content_analysis_model.py
chantisnake: you need to have doc string for every functions. Otherwise it will not pass the test we will implement in the future.

lexos/processors/analyze/content_analysis_model.py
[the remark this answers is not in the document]
chantisnake: I think what Emma means is what is the format of the content of the dictionary. Is it a yaml, a json, or just comma separated words?

lexos/processors/analyze/content_analysis_model.py
chantisnake: In this line you invoke 3 function that is nested in one another, and there is 5 nested brackets. This is line is messy, you may want to split it into 2 lines or add more comments

lexos/processors/analyze/content_analysis_model.py
chantisnake: @Landaluce We use keyword parameter by default unless it is uninformative, like `pd.Datafram(X=[[1,2],[3,4]])`, that `X` is uninformative. On this case, I would the keyword version would be `round(1.234325, ndigits=2)`; that `ndigits` is some what informative. Therefore I would agree with Emma.

chantisnake: also, typically we round only on the last step, because that is just formatting, not calculation. We want our calculation to be exact first so that we can make sure our test is correct. Here is one example: https://github.com/WheatonCS/Lexos/blob/master/lexos/models/similarity_model.py#L98

lexos/processors/analyze/content_analysis_model.py
chantisnake: - it is fine active_dict is not used. put this in the blackhole: `for i, _ in enumerate(active_dicts)` 
- give it a better name like `dict_index` 
- avoid working with index, it is really confusing

lexos/processors/analyze/content_analysis_model.py
chantisnake: I think Emma is referring to you `i`s and `j`s, not `xs`

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,470 @@
...
+    """Counts each phrase in the dictionary in the given file.
+
+    If a has with more than 1 word is found in a file, it is deleted
+    from the file to prevent double count.
+    For example:
+    dictionary = ["not very good", "very good"]
+    file = "not very good"
+    count: "not very good" = 1
+           "very good" = 0
+
```
chantisnake: actually you cannot put an empty line here

lexos/processors/analyze/content_analysis_model.py
chantisnake: can't you do this with regex? and you are only handling Ascii whitespace, this is against Lexos' definition of words

lexos/processors/analyze/contentanalysis_receiver.py
```diff
@@ -0,0 +1,47 @@
...
+    def __init__(self):
+        """The Receiver to get all the content analysis options"""
+        super().__init__()
+
+    def options_from_front_end(self) -> ContentAnalysisOption:
+        """Get the content analysis option from front end
+
+        :return: a ContentAnalysisOption object to hold all the options
+        """
+        options = [Option('calc_input', None),
```
chantisnake: There is truly no point in doing this dude. Also this is really insecure, every single option can possibly be `None`? Then you need to actually handle the `None` case in your code. 
This is not type-safe, and I think this code will not be able to pass mypy's check with strict mode on.

lexos/processors/analyze/contentanalysis_receiver.py
```diff
@@ -0,0 +1,47 @@
+from lexos.receivers.base_receiver import BaseReceiver
+
+
+class ContentAnalysisOption:
+    def __init__(self, formula, dict_label, toggle_all, dict_labels,
```
chantisnake: no type definition... at all?

lexos/processors/analyze/content_analysis_model.py
chantisnake: So, you can guarantee that the user will not be able to input a invalid equation?

lexos/processors/analyze/content_analysis_model.py
chantisnake: you can use indexes in panda data frame. That is how we handle headers.

lexos/processors/analyze/content_analysis_model.py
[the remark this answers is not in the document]
chantisnake: I think Emma means add documentation to this.
The rule of thumb is: when you have some line that require people to think, then add documentation.

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,488 @@
...
+        for index, (score, formula, file) in enumerate(zip(self.scores,
+                                                           self._formulas,
+                                                           self._corpus)):
+            scores_sum += score
+            total_word_counts_sum += \
+                file.total_word_count
+            formulas_sum += formula
+        if len(self.scores) != 0:
+            scores_avg = round(
+                (float(scores_sum) / len(self.scores)), ndigits=3)
```
chantisnake: in this function you are rounding couple of times. Is there a particular reason that you sometimes round to 3 digits, and sometimes round to 1? 
And since your are rounding to 1 digits couple of times, is it possible that you can create a name for it, to avoid it being a magic number?

lexos/processors/analyze/content_analysis_model.py
```diff
@@ -0,0 +1,488 @@
...
+    for phrase in dictionary:
+        count = 0
+        if file.content.startswith(phrase.content + " "):
+            count += 1
+        if file.content.endswith(" " + phrase.content + "\n") or \
+            file.content.endswith(" " + phrase.content) or \
+            file.content.endswith(
+                phrase.content):
+            count += 1
+        count += len(re.findall(_WORD_BOUNDARY_REGEX_STR + phrase.content +
```
chantisnake: all the previous cases can be handled just by expanding this regex.

lexos/processors/analyze/content_analysis_model.py
Landaluce: Yes, I round to 1 decimal place the averages of the raw counts for each dictionary, and to 3 decimal places the score (= formula result / total for count of the file) which is a smaller number. An alternative to rounding the score to 3 decimal places could be to convert scores to percentages.

lexos/processors/analyze/content_analysis_model.py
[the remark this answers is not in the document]
EmmaGrace: Yeah, pretty much. Also what the label is for

lexos/processors/analyze/content_analysis_model.py
EmmaGrace: I meant more that you could do a comprehension of sums_avg, total_word_counts_avg, etc, since these are always appended and I don't think the order they are in matters. You could then append your conditional averages to that

lexos/processors/analyze/content_analysis_model.py
EmmaGrace: I was referring to your whitespace-handling ifs. I agree regex will make this whole section much clearer

lexos/processors/analyze/content_analysis_model.py
chantisnake: In that case I think you need to make couple of constant for them to avoid being magic numbers

lexos/helpers/definitions.py
```diff
@@ -5,3 +5,20 @@
 _WORD_BOUNDARY_REGEX_STR = r'\s'
 
 WORD_REGEX = re.compile(_WORD_REGEX_STR, re.UNICODE)
+
+
+def count_phrase_in_text(phrase: str, text: str):
```
chantisnake: I think it would be nice if you can consider the fact that phrases starts or end with spaces.

lexos/helpers/definitions.py
```diff
@@ -5,3 +5,20 @@
...
+    """Count how many times the phrase appears in the text
+
+    :param phrase: string containing a phase or word
```
chantisnake: I would think a more precise definition of phrases will help future developers, like:
`a sequence of word that is separated by one space between every word` or `a string that ends on a word ending` or something else. 
You need to defining it based on your program, because this is a new terminology in lexos we need to make sure that this term is well-defined and well-documented for both of our developer and user.

lexos/helpers/definitions.py
Landaluce: Its done in the last line of the function: count += len(text.split(" " + phrase + " ")) - 1

lexos/helpers/definitions.py
```diff
@@ -5,3 +5,20 @@
...
+    :param text: string where the phrase will get count
+    :return: count: number of times the phrase appears in the text
+    """
+    count = 0
+    if text.startswith(phrase + " "):
+        count += 1
+    if text.endswith(" " + phrase + "\n") or text.endswith(" " + phrase) or \
```
chantisnake: I think it would be better if you can handle `"\r\n"`, although I think lexos currently converts line-ending on upload, I think it is nice to build more robust functions.

lexos/helpers/definitions.py
chantisnake: I mean phrase like " test test ", this will never get matched.

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,481 @@
...
+    :param dictionary: list of Phrase objects sorted by number of word in
+    descending order
+    :param file: a File object
+    :return: list of Phrase objects with their counts
+    """
+    for phrase in dictionary:
+        phrase.count = count_phrase_in_text(phrase=phrase.content,
+                                            text=file.content)
+        if ' ' in phrase.content:
+            file.content = file.content.replace(phrase.content, ' ')
```
chantisnake: I don't think this logic is correct... 

If I have phrase "not very good", and phrase "very   good", and I have test "very not very good good", then both phrases will be matched?

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,481 @@
...
+            error_msg += "tan takes exactly one argument (0 given)<br>"
+            is_error = True
+        if "log()" in self._formula:
+            error_msg += "log takes exactly one argument (0 given)<br>"
+            is_error = True
+        if is_error:
+            return error_msg
+        return ""
+
+    def analyze(self) -> [str, str]:
```
chantisnake: Type hinting is wrong. Should be (Optional[str], Optional[str])

chantisnake: Check if there is any other places where you need Optional in you type.

lexos/views/content_analysis_view.py
```diff
@@ -0,0 +1,213 @@
...
+                                                 dict_labels,
+                                                 active_dicts):
+            if active:
+                f = open(os.path.join(path, dict_name), "r")
+                content = f.read()
+                analysis.add_dictionary(file_name=dict_name,
+                                        label=dict_label,
+                                        content=content)
+        result_table, formula_errors = analysis.analyze()
+        if len(formula_errors) != 0 or result_table is None:
```
chantisnake: Formula error can be None, according to your documentation. Calling len function on None can generate error.

lexos/models/content_analysis_model.py
chantisnake: The "very good" has 3 spaces between them. Somehow GitHub removes the 2 spaces.

chantisnake:
My final comment to this pull request:

- After couple month I finally understand what each line does, therefore PLEASE ADD COMMENTS, EVERYWHERE. Please refer to any other branch for comment density.

- Here is some comment on the design of the code logic:

  - I personally think your definition of "Phrases" is non-intuitive, if the spaces between words counts as element of the phrase, then the phrase is more string-based than word-based, which is not what people would imagine a phrase should be. When you tell me the definition of phrases you also uses word to define it. Therefore I think the most intuitive definition of phrases is just "a list of word". This is just my personal opinion I don't know what others think.

  - The logical issue metioned [here](https://github.com/WheatonCS/Lexos/pull/660#discussion-diff-155152983R364), I think this can be categorize as bug. This problem is the same as to safely encode the place where a phrase is removed. There is absolutely no way to safely encode things unless you turn them into a data structure (object), or convert them to object notation (json, yaml, csv etc.). For example one of the safe way to encode the place where the phrase is removed is to store the index where phrases is removed along with the string (This is just an example, not the way to tackle this problem in this specific case). 

  - Unicode handling
  
  - One a very **HACKY** solution to solve both 2 problem above (definition, encoding issue, and unicode) is to remove all the Unicode space and sequence of space in phrases and text. There is a very hacky way to do this is `" ".join(text.split)`. However you will need to find the way to correctly manage the hack if you choose to go with this method (I personally would not hack), because the code is hacky it is very easy to get it wrong. Also please add abundant comments if you choose to go the hacky way. I can add more than 4 lines of comment for one single line of code if I think the code is not good practice, hacky and hard to understand.


I don't mean to be rude, but I have spend much more time than I should on this single pull request. 
I will just approve it for now, @Landaluce @EmmaGrace @Weiqi97  @mleblanc321 @scottkleinman , please work together to see when to merge this.

I have to apologize to Prof. Leibowtiz yesterday for not committing enough time on Honor Thesis. I really hate it. I work every single day in library from 10am to 10pm, and I just cannot find time for Honor Thesis.
Of course, this is not anyone's fault but myself. But I really need to spend more time on other things I committed to.

Thank you guys. I am really sorry, but I am afraid you guys need to take charge on this pull request.

lexos/views/content_analysis_view.py
Landaluce: I'm sue formula_errors is always a string, and clicking analyze without setting the formula should cause an error if formula_error was set to None, but it doesen't.

lexos/helpers/definitions.py
Landaluce: why would anyone put " test test " instead of "test test" in the dictionary?

lexos/helpers/definitions.py
Landaluce: I can do that, but isn't redundant if lexos already handles it?

lexos/models/content_analysis_model.py
Landaluce: Again, why would anyone put "very   good" instead of "very good"? Also, the only way I can think of fixing this is replacing the found phrase with some kind of tag, but there is always the possibility of that tag been in the dictionary anyway.

lexos/views/content_analysis_view.py
Landaluce: I just fixed the documentation

lexos/helpers/definitions.py
chantisnake: @Landaluce that is the point of robustness, you don't want your function only to work on a specific type of input. If `" test test"` is a valid phrase based on your definition, then it should be handled.

lexos/helpers/definitions.py
chantisnake: robustness, always about robustness

lexos/models/content_analysis_model.py
chantisnake: same as https://github.com/WheatonCS/Lexos/pull/660#discussion-diff-155119410R10

lexos/helpers/definitions.py
chantisnake: You don't need to fix this right now, you can just leave them in the documentation, saying some specific cases will not be handled by this function

chantisnake:
Clarification on my previous statement, copied from my conversation with Mark:
```
That message is a pure FYI for you guys, if you guys think those issue don't need to be sorted out, I totally understand, people want to use this tool. 
If those issue is not fixed before it is merged, we can put those message in issues and documentation, so that future developer can look at them.
```

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,481 @@
...
+        """Generates a dataframe containing all values stored in this class.
+
+        :return: a data frame containing all values stored in this class
+        members
+        """
+        columns = ['Document Name'] + [dictionary.label for dictionary in
+                                       self.get_active_dicts()] + \
+                  ['formula', 'total word count', 'score']
+        indices = [file.label for file in self._corpus] + ['averages']
+        df = pd.DataFrame(columns=range(len(columns)),
```
Weiqi97: Is it possible that instead of working with **col_index** this Dataframe can work with actual column names? 
`columns=["file_label", ...]`

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,481 @@
...
+                          index=range(len(indices)))
+        # adds row to df with: file label, count for each dict,
+        # formula result, total word count, score
+        for row_index, (file, formula, score, counter) in enumerate(
+            zip(self._corpus, self._formulas,
+                self._scores, self._counters)):
+            df.xs(row_index)[0] = file.label
+            for col_index, (count) in enumerate(counter):
```
Weiqi97: Some documentation here for this for loop might be nice!

lexos/models/content_analysis_model.py
```diff
@@ -0,0 +1,481 @@
...
+        test_option:
+        dictionaries: List of Dictionary objects
+        corpus: List of File objects
+        counters: a 2D array with count of every dictionary for evey file in
+        the corpus
+        formulas: List of string that represent a formula for each file
+        scores: List of formula/total word count of each file
+        averages: Lis of averages count of each dictionary
+        """
+        self._test_options = test_options
```
Weiqi97: Maybe some simple explanation for these variables will be helpful?

Weiqi97:
Since I don't know much about **regex** and I'm not very familiar with dealing the actual texts, I can't really help with the problems we've discussed in recent two meetings. Sorry about that. (I guess **Dataframe** is my thing now!)
By now we have all worked on flatting existing code and converting models etc... One reason I think we committed to do this instead of adding much more awesome new functionalities is that we want to keep Lexos develop-able. Lexos has been out there for years and it definitely will survive longer. What we are doing now will benefit the coders who walk in after us. And what we wanted is that when others walk in, they don't have to spend hours to read our code just to fix a simple bug. So I think more **documentations** will always be nice.
We may feel the same that once we worked on some code for long, every single line becomes very easy to understand by ourselves. And thus, adding more documentations probably will not be as simple as it sounds like. However, future coders who have absolutely no experience on working on such a large project will need those documentations. I think one place we may always want to add more documentations is where we could not possibly avoid nested for loops...
Also, adding comments to places where could possibly exist a better solution is also a nice thing to do. As we discussed, we can feel safe to add **TODOs** for things we may want to do but did not have enough time to do.
And now I really have to go work on **Topword** and add my documentations...

lexos/models/content_analysis_model.py
Landaluce: If we do so, how can we then do these lines:
            for col_index, (count) in enumerate(counter):
                df.xs(row_index)[col_index + 1] = count
            df.xs(row_index)[col_index + 2] = formula
            df.xs(row_index)[col_index + 3] = file.total_word_count
            df.xs(row_index)[col_index + 4] = score

chantisnake: df.concat

chantisnake: Sorry, df.append.

PR #671, opened 2017-11-20
Feature/similarity model

chantisnake:
## backend:
- implement the similarity model, remove the old similarity function
- rename one of the file manager function to a more sensible name
- change the old return from "2.3***3.4" to an html table generated by panda
- shorten the view function

## frontend:
- clean up the frontend
- change the frontend display to dataTable
- move the csv download functionality to be handled by dataTable

## test
- remove the old similarity test
- change the test to test the result series


lexos/receivers/similarity_receiver.py
```diff
@@ -0,0 +1,24 @@
+from typing import NamedTuple
+from lexos.receivers.base_receiver import BaseReceiver
+
+
+class SimilarityOption(NamedTuple):
```
Weiqi97: Are we changing this to SimilarityFrontEndOption?

chantisnake: Yes

PR #674, opened 2017-11-30
update word boundary regex to be more precise

chantisnake:
more precise definition about word boundaries


chantisnake:
closed because we need more discussion about this subject.

See slack for detail.

PR #686, opened 2018-05-16
remove JetBrains files

Weiqi97:
#### Removed all files under .idea folder and make GitHub not sync this folder anymore.
  - The reason is that project configuration files (as well as other files that specifies interpreter location) differ from platforms or based on our installation of the interpreter.


Weiqi97:
Here is one possible way to solve the problem. We can enforce everyone to rename the interpreter to something that doesn't include the full path, as the project then will then store name of the interpreter. So the configuration files might be able to be shared among platforms. But I'm not sure if this is worth doing.

chantisnake:
No it is not, but I do encourage you to keep code style setting.

Weiqi97:
@chantisnake That sounds good. I will make sure I exclude the code style settings. And sorry instead of reverting the changes.. I will create a new branch and only delete the files we don't want to sync.

chantisnake:
I believe three task configurations can also be platform independent. I will try to take a look

chantisnake:
I found this document which may of interest to you.

Weiqi97:
Yes I think that will work. That's what I mentioned above that we enforce everyone to name their interpreter the same way. But I thought that wasn't worth doing...

chantisnake:
My edits are done, however I noticed git sometimes have trouble ignoring `Lexos.iml` and `misc.xml` I have no idea why...

chantisnake:
Turns out ignoring further changes in file are fucking hard to do: https://stackoverflow.com/questions/4114163/how-to-tell-git-to-ignore-all-further-edit-to-a-single-file-without-removing-it

I would suggest to investigate the multiple interpreter approach. (Add many interpreter and let people choose which one corresponds to their system

chantisnake:
I think conda environment would be a good solution to this (documents can be found here: https://conda.io/docs/user-guide/tasks/manage-environments.html#create-env-file-manually)

I want to let people create a conda environment in an OS independent directory (say `$ProjectDir` which is the lexos directory: `/stuff/like/this/lexos/`).  
However we need to enforce where lexos should be cloned into. In the wiki we recommend to clone lexos on desktop, but I personally think `$HOME/githubRepos` or `~/githubRepos` would be a better place. What do you think?

---
By the way, there is an error on the wiki `lexos bootcamp` page that let people install `python-pip` and then run `pip` install. It should be just `python -m pip install -r requirement.txt`, this command will correctly use anaconda's pip to install package. 
If `python-pip` is installed, the system may load python's pip instead of anaconda pip.

I have fixed it.

mleblanc321:
oh, good; thanks for fixing that, cheng (you are a faithful worker!)

.idea/runConfigurations/lexos_core.xml
```diff
@@ -6,9 +6,9 @@
     <envs>
       <env name="PYTHONUNBUFFERED" value="1" />
     </envs>
-    <option name="SDK_HOME" value="" />
+    <option name="SDK_HOME" value="$PROJECT_DIR$/../anaconda3/bin/python" />
```
chantisnake: I think you want to use module sdk here

.idea/runConfigurations/test.xml
```diff
@@ -1,27 +1,18 @@
...
-    <option name="PATTERN" value="" />
-    <option name="USE_PATTERN" value="false" />
-    <option name="testToRun" value="" />
-    <option name="keywords" value="" />
-    <option name="params" value="" />
-    <option name="USE_PARAM" value="false" />
-    <option name="USE_KEYWORD" value="false" />
+    <EXTENSION ID="PythonCoverageRunConfigurationExtension" runner="coverage.py" />
+    <option name="_new_keywords" value="&quot;&quot;" />
+    <option name="_new_additionalArguments" value="&quot;&quot;" />
```
chantisnake: I don't quite understand the following 3 lines. Please review and make sure they are okay.

Weiqi97: I think they are fine. There were keywords and additional arguments settings in the pytest and I left them empty.

.idea/runConfigurations/lexos_core.xml
Weiqi97: I changed the interpreter settings to project default, I hope it will do the trick.

PR #692, opened 2018-05-21
Bugfix/block titles renamed

danny98m:
## New Naming for Block Titles
The naming for block titles was inconsistent before so I have made it all uniform as it now displays the name of the current tool.


chantisnake:
- this pr contains way more change than just renaming block title
- I may be wrong, but as far as I remember the block title are supposed to be 'analyzer' 'visualizer' and 'manage'. Only context analysis has the wrong title, other tools are correct.

mleblanc321:
huh?   danny asked me today and i made the decision that each tool would be noted by its role; is this the objection?  that is,  not "analyzer" but "topWords", etc.

chantisnake:
Uh, okay. That make sense. But i think instead of 'top_word', 'TopWord' is better. Because a tool resembles a class rather than a variable.

mleblanc321:
oh, really?  we were thinking of variable name rather than class, but you are right;  danny, thoughts?

scottkleinman:
This is for the front end, right? It sort of makes sense to use Javascript styling conventions (typically camel case). But that might be too fine a point. Why not ask some non-coders what looks best/most meaningful to them?

That said, I agree that we should give the title of each tool. Maybe we can eventually adopt something like a breadcrumb strategy to show the top-level menu items, but right now I don't think we've got the screen real estate.

danny98m:
I agree. I think asking some of the users what they think looks better is really the best idea since there opinion is probably the most important when it comes to pleasing visuals.

mleblanc321:
danny: the English students show up today in larger numbers;  how about if you poll them individually, showing them options;  include me too :)

danny98m:
Sounds like a plan!

Weiqi97:
@danny98m Please discuss the behavior of error messages too if possible. i.e. should we display the error message when they got to the page or only display the message when they click on generating a result. Thanks.

danny98m:
I'm going to briefly mention it tomorrow morning with the English users and whoever else wants to participate to get some sort of agreed upon standard real quick. After that's decided I'll make the changes and submit a new pull request. I agree that camel case looks the best and makes the most sense.

danny98m:
I provided the English students with 4 options (camelcase, underscores, lowercase separated by space, and uppercase separated by space) and the majority selected uppercase separated by space. None of them seemed to like camelcase or underscores.

This file contains two examples of each proposed option: [Block Titles Design Options.pdf](https://github.com/WheatonCS/Lexos/files/2032607/Block.Titles.Design.Options.pdf)

scottkleinman:
The history behind this is that we thought we were being very clever in using "code-like" curly brackets. But it seems that our users feel more comfortable with natural language formatting...

jaronson13:
oh well thats a tricky situation then because if it's the students who are requesting the changes then i'd say go with what they think is  best for the ui design, but i also think the changes are not too large to make a big difference in appearace, rather the users who use it daily might just prefer the slight change. But if you ask users who use it ever  so often how they think it looks then they might have a different option for how the titles are displayed vs how it looks to regular users (like the english students and others in the field) . It would be great to see some options from use who dont use it often as english majors but just people awho might have this repo starred/forked, Any idea on the traffic we get from our current stable version URL? maybe we could ask users to do a quick survey regarding the visual layout

jaronson13:
yes im on mobile so i apologize for the bad grammar. im just checking in to see how this project is doing and seeing if i can review anything, specifically ui elements. I just looked at the ui after not seeing it for a while and i can tell there are alot of good changes, and i like the margins as well. But I agree that the titles should be changed  to something else. I like the idea of camel case

chantisnake:
I suggest you to delete this branch and start a new branch to work on this. Remember not to commit files in `.idea`, and all the other js and css files.

danny98m:
@chantisnake Yeah my bad. I have already started another branch for this. I just wasn't sure if deleting this current branch would close this thread for messaging. When I made this pull request I clearly still wasn't too confident with github so I accidentally committed things I didn't mean to. I understand how to do it all now though haha. I'm going to hold off on submitting a new pull request for this topic until it's finally agreed upon what format we should use for the block titles.

scottkleinman:
Jesse's suggestion is a good one, but most of the followers on GitHub are former Lexos developers, and we haven't had too much success getting survey responses on Twitter. I'd vote for just going ahead with natural language formatting to move on. That way, we'll have solved the consistency problem, and we can make changes globally later if we want to.

chantisnake:
I don't think close this thread will close the messaging. It would be nice if you reference this pr (paste the url of this pr) to the new pr, and say that we made the decision blablabla, for more, see https://github.com/WheatonCS/Lexos/pull/692

PR #694, opened 2018-05-23
Set cutter last segment size boundary

mathewleblanc:
Added a max percent value for the last segment size threshold. 
 - The max is now 100 and the min is still 0.


chantisnake:
No, no, no, sorry, I don't mean to approve this...

chantisnake:
To the best of my understanding, current cutter can handle more than 100%. Yeah @Shi26 !
However, I don't know if that is a reasonable for the user to use this feature. 

Please read this function to understand it:https://github.com/WheatonCS/Lexos/blob/master/lexos/processors/prepare/cutter.py#L9
**This function is super hard to read**... We try to make it easier, but failed.

I recommend to ask the English people about whether they want to have last prop more than 100%...

Weiqi97:
Since we decided that we should set a boundary so that users do not go over 100% for the last segment size, can we add a tool tip to explain that we are doing this?

PR #702, opened 2018-05-24
JS 2.x Events Update

mathewleblanc:
Updated all of our jQuery 2.x events such as `.click(function(){})` to `.on('click', function(){})`. The shorthand syntax is deprecated in 3.x so the longer `.on` should be used.

This version is using 3.3.1. I used this [migrate](https://github.com/jquery/jquery-migrate/#README) plugin to locate compatibility issues which are logged in the console. It's logging a warning about the CDN tables which I am unsure of how to fix.

I have not yet changed any of the Ajax calls.


chantisnake:
Please provide deprecation notice for these API's.

mathewleblanc:
Here is one example of the output in the console from the plugin:

shot1

chantisnake:
I was looking at the documentation for `click` there is no deprecation notice https://api.jquery.com/click/

I also search through the deprecation list https://api.jquery.com/category/deprecated/ `click` is not in there. There is only `bind` and `delegate`, which is the same as `on`.

I think we do need an official deprecation notice to remove all the `click` function.

mathewleblanc:
If that is the case I am not sure why the plugin is giving these warnings. It does seem odd that it would even though it does not appear to be officially deprecated.

chantisnake:
It is indeed curious, since jquery-migrate is made by jquery.

You can just post an issue on their repo and ask them about it.

scottkleinman:
The discussion on the [jQuery repo](https://github.com/jquery/jquery/issues/3214) was pretty convoluted as of last November. This may be a case where the migrate plugin may have got ahead of what is actually happening.

mathewleblanc:
Thank you, that definitely makes a lot more sense. Tomorrow I will create a new branch using 3.3.1 without changing anything else.

chantisnake:
It seems like they are trying to remove it, however i cannot understand why.

typically people deprecate redundant, low-level, out-dated api. `click` is neither of them. Nevertheless we are using jquery, and I am not in charge of jquery development.

I am probably gonna offend lots of people, and sounds incredibly arrogant, but I am just gonna go ahead and say it. Freaking string is not safe, is hard to work with and makes code analyzer confused. jQuery people should learn.

Sigh...

chantisnake:
I believe if you unchange all the files in the `.idea` folder we can merge these changes.

PR #706, opened 2018-05-25
manage/ajaxCall_update

phuens:
Updated the Ajax call for the manage.js


Weiqi97:
Closed because it was depend on another developing branch.

PR #710, opened 2018-05-25
Ajax call update for manage page

phuens:
updated the Ajax Call


chantisnake:
Please move all the css and html changes to a separate pull request. Thanks!

PR #718, opened 2018-06-01
Update .editorconfig

Weiqi97:
### Change js indent size to **2**.
We may want to change this since this file will overwrite the editor settings in PyCharm. Alternatively we also can tell PyCharm not to read this file..


chantisnake:
uh, thank you. I think this might be my fault...

PR #723, opened 2018-06-04
Refactoring of Cutter's Front End (JS and HTML)

danny98m:
# Refactoring of Cutter's Front End (JS and HTML)
* removed var in favor of const and let.
* Used strict equalities where preferred. 
* Broke up functions into sub functions. 
* Changed variables to camelCase naming standard.
* Changed strings utilizing concatenation to template literals.
* Removed unused parameters and variables. 
* Cleaned up html (removing unused tags)
* Fixed typos in HTML.


Weiqi97:
It would be nice if you can add your `scripts_cut.js` to eslint.

Please see this: 900b09c

danny98m:
Should I do that within the travis and appveyor files? Or just run eslint in my terminal?

Weiqi97:
Add to both .travis.yml and appveyor.yml. See commits mentioned above for syntax details.

danny98m:
Ohhhh I see. My bad. Will do that right now.

Weiqi97:
Please view [this page](https://eslint.org/docs/rules/valid-jsdoc), if you need extra information about JSDoc.

danny98m:
#### I have fixed all the eslint errors except for the following:
*  130:20  error  'activeFileIDs' is not defined         no-undef
*  144:43  error  'numChar' is not defined               no-undef
*  147:48  error  'numWord' is not defined              no-undef
*  149:48  error  'numLine' is not defined                no-undef
*  167:14  error  'numChar' is not defined               no-undef
*  174:14  error  'numWord' is not defined              no-undef
*  180:14  error  'numLine' is not defined                no-undef
<hr>
These are errors simply because these variables are actually defined within cut_view.py. So it is not a problem. Eslint just won't recognize these as defined because obviously it isn't looking at cut_view.py.

Weiqi97:
I'm not familiar with this file.. But if you haven't tried before, could you please try to move the variable declarations in ```cut.html``` to ```script_cut.js```? And if the value does not need to be changed, please change them to ```const```.

Weiqi97:
[This](https://stackoverflow.com/questions/21626048/unable-to-pass-jinja2-variables-into-javascript-snippet) might be helpful for moving the variables...

Weiqi97:
@danny98m Danny could you please try to do the same as what you did in the error message branch? (Use meta to store the values u need.)

chantisnake:
These should be ajaxed, I believe. Can you figure out a way to change them into Ajax?

danny98m:
This file now pleases ESLint

danny98m:
Moving the on click process function all into the js (currently being called from onclick attribute in html elements) is causing a lot of issues for some reason.... that is why eslint is failing once again.

lexos/static/js/scripts_cut.js
```diff
@@ -1,277 +1,401 @@
...
+    This may be because you are cutting a large number of documents. 
+    If not, we suggest that you cancel, reload the page, and try again.`
+    const footerButtons = `<button type="button" class="btn btn-default" data-dismiss="modal">Continue Anyway</button>
+    <button type="button" class="btn btn-default" id="timerCancel" >Cancel</button>`
     $('#warning-modal-footer').html(footerButtons)
     $('#warning-modal-message').html(timeWarning)
     $('#warning-modal').modal()
-  }, 10000) // 10 weconds
+  }, 10000) // 10 seconds
   xhr = $.ajax({
```
Weiqi97: I think we may don't need ```xhr = ``` for the ajax call. The documentation can be found [here](http://api.jquery.com/jquery.ajax/). Maybe try to remove it and see if it still works? Thanks..

danny98m: Yes still works fine after removing xhr.

PR #725, opened 2018-06-04
jQuery 3 Update

mathewleblanc:
# jQuery 3 Update

This is now using jQuery 3.3.1. The 2.2.4 script tag in base.html has been replace with 3.3.1. The features in each tool have been tested and so far appear to be functioning normally. I do not believe that any of the functions, shorthand etc have been deprecated. Hopefully there is not much else we have to change.


lexos/static/js/scripts_viz.js
```diff
@@ -174,7 +174,7 @@ $(window).on('load', function () {
       canvas = document.getElementById('svg-canvas')
 
     // Draw svg on canvas
-    canvg(canvas, content)
+    //canvg(canvas, content)
```
chantisnake: please explain why this code is commented out.

chantisnake: I see your comment saying that it is unnecessary. I am not familiar with this code. Maybe Scott knows why?

mathewleblanc: For now we can ignore this. It was just something I noticed that is unrelated to the jQuery update. This can be something that someone investigates more when this page is refactored.

scottkleinman: It looks to me like this is a remnant of a former method of rendering a .png file using `canvg.js`. I would revise the code to this:

```javascript
  // Save to PNG
  $('#save').on('click', function () {
    /* Code may be obsolete/deprecated -- check to see if a PNG is generated without canvg.js */
   /*
    var $container = $('#viz'),
    // Canvg requires trimmed content
    content = $container.html().trim(), // Note trailing comma here. Broken code?
    // Draw svg on canvas
    canvg(canvas, content)
    */

    canvas = document.getElementById('svg-canvas')

    // Change img from SVG representation
    var theImage = canvas.toDataURL('image/png')
    $('#svg-img').attr('src', theImage)

    // Open a new window with the image
    var w = window.open()
    var img = $('#svg-img').clone().css('display', 'block')
    var html = $('<div/>')
    html.append("<h3>Use your browser's Save as function to save the image</h3>")
    html.append(img)
    $(w.document.body).html(html)
    // End Save
  })
})
```
And perhaps open up an issue to check the PNG generating function.

Weiqi97:
I gonna merge this change now.. Thus we got a bit more time testing this while we are developing..

Issue #611, opened 2017-08-04
Misc tool tip problems


EmmaGrace:
For "Statistics > Assign Temporary Labels", the tool tip icon does not show the tool tip.

In "Kmeans > K-Means Options > Method of Visualization", the tool tip disappears before the user can click on the links.

In "Kmeans > Culling Options", the drop down arrow does not turn.

In "Similarity > Select Comparison Document", the text formatting is inconsistent and the text box not correctly sized in the floating window.

In "Similarity > Culling Options", the drop down arrow does not turn.

In "Topword", the placement of "Assign Temporary Labels" is inconsistent.

J

chantisnake:
Is this one fixed? I remember there is a pr about tooltips merged into master.

EmmaGrace:
If you are referring to the pull request "bugfix/documentation_grammar" that Josh made using my account, that was merely grammar fixes. The above is everything we couldn't change during that pass because it requires actual code and not just string changes

chantisnake:
I think this is a front end change. Do we have people to look at this right now?

mathewleblanc:
On local host, the tool tip for "Statistics > Assign Temporary Labels" seems to be working. "Assign Temporary Labels" in Topword is still slightly misplaced compared to the other pages. I will try to fix this now as well as the culling options drop down arrow.

jacksonjreed:
only the fourth issue is still present

PR #720, opened 2018-06-01
feature/stats_model_update

Weiqi97:
## Update statistics model
- Add a new box plot for statistics of the entire corpus.
  - It helps user easier identify outliers.
- Make most functions to use pandas instead of NumPy.
- Updated Ajax calls to easier display data table and box plot.
  - Let back end directly return a HTML format table.
- Fix file toggling bug in current master.
  - Toggle individual file on or off at statistics page was not working on current master branch.
- Some naming fix and updated unit test.

PR #729, opened 2018-06-05
Chevron Button Animation Fix

mathewleblanc:
# Chevron Button Animation Fix

In response to #611, I fixed the animations for the chevron button drop downs. I am still not sure why the tooltip for Assign Temporary Labels in K-Means is not working but at least the rotate animation should be working now.


Weiqi97:
I'm not sure about this... But maybe instead of repeating the functions, try add the function to ```scripts_analyze.js```.

Weiqi97:
I think once #720  is merged.. We will work on a refactor of ```base_analyze.html``` and ```scripts_analyze.js```. There are duplicated functions in a couple different tools, we may move them as well.

Issue #696, opened 2018-05-23
Make presence of glyphicon-chevron-right more obvious to user.


danny98m:
## Glyphicon Visibility
Talking to users, I have discovered that the little arrow on the right hand side of the page for each tool is very hidden. Most users seem to not even notice it is there until you point it out. The information provided by clicking the glyphicon is really helpful so I think it would be beneficial to make the tab move or something when a user initially goes to a page. Just to attract attention to it. This would be some pretty simple jQuery just not sure what people want the attracting action to be. 

### Proposed Enhancement:
* Add some sort of action to that pull out info tab so that its presence is more obvious to the user.

jaronson13:
yes that would  look better i think for the ui pages, mainly for ease of use. Other things i noticed are the in the marghins pages can be hard to see where to close. The arrow is not always easy to see and notice

mleblanc321:
_ok, who is spoofing on the jaronson13 account?  (cheng?)_

mleblanc321:
phuntsho will show us a demo of one possibility ...

chantisnake:
I am definitely not spoofing on @jaronson13's account. lol

danny98m:
I have what I believe is a good solution to this. I'll work on it some more after some other tasks.

danny98m:
I made it so the sidebar (In  the Margins) pops out for a little and then returns to the original space to make user aware of it. This only occurs on the upload page.

Here is the relevant pull request [#701](https://github.com/WheatonCS/Lexos/pull/701)

jaronson13:
yes i think the concept is good, just hard to know how users would respond to it. I think it makes me aware of it but maybe non-technical user would prefer to have a tour like @scottkleinman  mentioned. definitely worth discussing further

danny98m:
Okay, I can talk about it with the English students if that seems sufficient enough?

chantisnake:
This issue is addressed in pull requests: https://github.com/WheatonCS/Lexos/pull/701

danny98m:
I personally feel web storage is a better way to do this than using persistent cookies as it is more secure and you can fit more data. I have an alert popping up based on if a user has visited before or not. Its just displaying a test message as I'm not sure what we would want the alert to say or if an alert is even what we want.

scottkleinman:
[...]

PR #730, opened 2018-06-05
Upgraded Sidebar Presence

danny98m:
## Sidebar Presence Now Made Known Through Popover
The first time a user uses Lexos a popover will appear alerting the user of the in the margins tool. I'm using local web storage so this will only happen once unless the user clears their browser's cache. 
Here is a terrible quality gif:
sidebar


danny98m:
Here is the relevant issue [#696](https://github.com/WheatonCS/Lexos/issues/696).

scottkleinman:
The popover could use some styling (maybe wider), but otherwise I think this implementation is good. You could possibly reset local storage to remove 'visited' when the session is restarted, but I am undecided as to whether that's a good idea.

danny98m:
I added some style to this popover. If you would prefer session storage rather than local storage let me know and I'll change it.

mleblanc321:
@danny98m - give me a demo of this and we'll discuss

lexos/static/js/scripts_base.js
```diff
@@ -1,3 +1,23 @@
+$(function () {
+  if (localStorage.getItem('visited') !== 'yes') {
```
Weiqi97: Great work! One last thing is that.. Could you please attempt to create a function for this part and call the function in the existing: ```$(function () {})```. Add if the function works, please add a doc string for it.

lexos/static/js/scripts_base.js
```diff
@@ -1,3 +1,23 @@
...
+    localStorage.setItem('visited', 'yes')
+    $('#toggler').popover({
+      'html': 'true',
+      'title': 'In the Margins!',
+      'content': 'View instructions for any screen in Lexos by clicking the In the Margins Tab!\n<div class="text-center"><button type="button" id="gotit" class="btn btn-primary" style="background-color: #0068AF">Got it!</button></div>'
+    }).popover('show').data('bs.popover').tip().css({
+      'width': '170px',
+      'text-align': 'center'
+    })
+    $('#gotit').on('click', function () {
```
Weiqi97: I don't really remember if we agreed on using ```.click``` or ```on('click')```..

lexos/static/js/scripts_base.js
danny98m: Oh yes my bad I should have done that before after testing but forgot. Just did this.

lexos/static/js/scripts_base.js
danny98m: I believe we actually agreed on .click. Just changed it.

Issue #501, opened 2017-01-24
Truncate long document labels


scottkleinman:
Long document labels can overflow their container elements, so it is advisable to truncate them. This has already been done in Upload, Scrub, and Cut, but it is advisable to do so throughout the UI. In Jinja, this is easy:

```python
{{ documentLabel|truncate(5, True) }}
```
This will output `docum...`. The exact number of characters used for truncation will depend on the width of the container element.

In some cases, it may be appropriate to do it on the back end (e.g. `documentLabel[0:5]`, or on the front end with css like the implementation in `upload.js`.

chantisnake:
I am no expert, but this seems like a issue that can be solved by text wrapping.

scottkleinman:
In most cases, document Labels would look a little odd if wrapped, so truncation is a better approach. It can be done easily in CSS with

```css
{
  width: 90%;
  margin: auto;
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
```
This will pace the label in centre of its container element and truncate the text with a trailing "...".

chantisnake:
Sounds reasonable

chantisnake:
Weiqi, send someone to investigate this please. I don't think this is an hard fix.

mathewleblanc:
I am working on this now. So far I have truncated the document name and original source on the manage page.

mleblanc321:
yes, i believe we want this in tokenizer also;  @mathewleblanc -- careful as @Weiqi97  has done some great work already on tokenizer (@mathewleblanc please pursue in coordination with weiqi)

mathewleblanc:
I believe #732 fixed this and we can now close this issue.

PR #732, opened 2018-06-06
Document Name Truncation

mathewleblanc:
# Document Name Truncation

Here is a fix for issue #501. I used a combination of Jinja and CSS to truncate long document labels so they look better in the UI. This has not been added to tokenizer yet as Weiqi is doing some major work on that tool. I will work in his branch and add it there.


Weiqi97:
Nice work!

Issue #520, opened 2017-02-25
Browser detection


scottkleinman:
Our Windows users are more and more likely to use IE/Edge, and our Mac users Safari. It would be nice if we could support these browsers, but our limited resources prevents it. So it may be worthwhile to use some basic browser detection to provide a warning if the user is not using Chrome or Firefox (or is using a mobile OS for that matter). I have a short Javascript that does the trick, but how and where to provide the warning is a little tricky.

Down the line, it might be worthwhile to enhance our browser support using [Modernizr](https://modernizr.com/).

mathewleblanc:
@scottkleinman do you still have the Javascript you mentioned? If not I believe I have a script that can detect the user's browser, but like you said, we must choose a good place and manner to display the warning. Perhaps we can do something similar to @danny98m in issue #730. If we want the warning to appear only once, we may want to also use web storage and display it when the user reaches the upload page. If we want to display it every time then we won't have to use web storage or any such tracker.

scottkleinman:
I think something like this should do the trick:

```javascript
function BrowserDetection() {
    //Check if browser is IE
    if (navigator.userAgent.search("MSIE") & > = 0) {
        // show warning
    }
    //Check if browser is Chrome
    else if (navigator.userAgent.search("Chrome") > = 0) {
        // pass
    }
    //Check if browser is Firefox 
    else if (navigator.userAgent.search("Firefox") & > = 0) {
        // pass
    }
    //Check if browser is Safari
    else if (navigator.userAgent.search("Safari") &> = 0 && navigator.userAgent.search("Chrome") & < 0) {
        // show warning
    }
    //Check if browser is Opera
    else if (navigator.userAgent.search("Opera") & > = 0) {
        // show warning
    }
}
```

mathewleblanc:
Thank you, that will be very useful. Should we use local or session storage to know when to display the warning? Local storage will be much less invasive as after the first time the user reaches the upload page the warning will never be displayed again unless they clear their history/cache. However, session storage might be nice because it will let them know that they are using an unsupported browser every time they start a new session. This would be good in case the user forgets that their browser is not supported. I am more in favor of session storage but I am not sure if this might be so frequent that it may annoy the user.

mleblanc321:
session storage seems to be the right choice (?) for this, since using an unsupported browser should be shared (here, on each reset)

chantisnake:
I believe there will be an npm package for this. Please don't reinvent the wheel.

[...]

PR #734, opened 2018-06-06
Browser Detection

mathewleblanc:
# Browser Detection

In response to issue #520 I have added browser detection to the upload page. The user's first visit to the page is saved as a boolean in session storage. If they are using an unsupported browser (anything other than Chrome and Firefox) an alert appears with an error message informing them that their current browser is not supported. This alert will not appear again when the user visits the page until the current tab is closed and they visit again or they clear their cache/history. This should be a decent temporary fix until we decide to use Modernizr as Scott suggested.


lexos/static/js/scripts_upload.js
```diff
@@ -194,4 +194,51 @@ $(function () {
...
     Init()
   }
+
+  browserDetection()
 })
+
+function browserDetection () {
+  // Display a warning if using an unsupported browser
+
+  var browserWarningMsg = `You are currently using an unsupported browser.\
```
Weiqi97: Please use ```const``` if this variable does not get changed. Same for the couple declarations below.

lexos/static/js/scripts_upload.js
```diff
@@ -194,4 +194,51 @@ $(function () {
   if (window.File && window.FileList && window.FileReader) {
     Init()
   }
+
+  browserDetection()
 })
+
+function browserDetection () {
```
Weiqi97: Please add a documentation here for this function.

lexos/static/js/scripts_upload.js
```diff
@@ -194,4 +194,51 @@ $(function () {
...
+
+  // Chrome 1+
+  var isChrome = !!window.chrome && !!window.chrome.webstore;
+
+  // If first time on page, show error message if browser is not supported
+  if (typeof(Storage) !== "undefined") {
+    if (!sessionStorage.visited){
+      if (isOpera) {
+        alert(browserWarningMsg)
+      }
```
Weiqi97: Use the Eslint forced style:

```javascript
if ( ) {
  // Code
} else if ( ){
  // Code
} ...
```

chantisnake:
Don't write your own code use the npm package.

chantisnake:
Browser detection will be hard to maintain. We have to use a package for this.

mathewleblanc:
All of the packages I can find with npm use `require()` which is not supported in default Javascript but is supported in Node.js. If anyone can find a package which does not use require that would be great.

chantisnake:
`require` is not only supported in node. You can just use import. Or you can just include them in HTML and use the function in the file.

mathewleblanc:
I'm trying to use `import 'detect-browser'` but I keep getting this error:
`SyntaxError: import declarations may only appear at top level of a module`
I can't find much information on how to fix this. I think `import` is definitely the best way to do this but I do not know how to resolve this. [This](https://www.npmjs.com/package/detect-browser) is the package I am trying to use

chantisnake:
I think the error message is pretty clear. You need to write import at the beginning of a file.

mathewleblanc:
That is exactly what I did. It did not work.

chantisnake:
What does Google tell you, my friend.

PR #738, opened 2018-06-07
Frontend Refactoring of Topwords

danny98m:
# Refactored Topwords 
This one was mostly good to go. Just moved functions, fixed spacing, fixed implicitly declared variables, removed some unnecessary stuff, and added documentation.


Weiqi97:
We are trying to add the ajax call to this tool since this is the only one left. And once the work is done, we will open a new pull request.

PR #741, opened 2018-06-08
Analyze/Chevron Animation Fixes

mathewleblanc:
# Analyze/Chevron Animation Fixes

I cleaned up base_analyze.js so it now meets ESLint standards. All but one instance of the chevron rotate animation has been fixed. I am not sure why the chevron for assign temporary labels in K-Means is not working. If someone else could take a look at this I would appreciate it. The problem before was the animation Javascript was repeated in multiple files, conflicting with itself when toggling the classes for the proper CSS rules. This script is now shorter and only in base_analyze.js


Weiqi97:
Good work! Sorry I forgot to tell you, since we removed gray word feature, can you remove if from the front end as well? Thanks.

mathewleblanc:
I just removed grey words front the front end. It was only in base_analyze.html and scripts_tokenizer.js.

PR #742, opened 2018-06-08
Feature/pydocstyle

Krissac:
finished all models except stats model


chantisnake:
Really nice job. I look forward to adding pydocstyle to the CI.

PR #743, opened 2018-06-08
Remove Redundant Ajax in  Dendrogram/Similarity

mathewleblanc:
# Remove Redundant Ajax in  Dendrogram/Similarity

This just removes a redundant `sendAjaxRequest()` from scripts_dendrogram.js and scripts_similarity.js and moves it into scripts_analyze.js. The Ajax still functions properly either way but there is no need two repeat it in these 2 separate files.


chantisnake:
I think we should discuss whether to use import. Because that will make the code analyser in pycharm and eslint understand our code better, and make integrating webpack a lot easier. This feature is supported in most modern browser. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import

Weiqi97:
This change will be in #746.

PR #746, opened 2018-06-11
Base_analyze Refactor

mathewleblanc:
# Base_analyze Refactor

- Chevron button rotate animation fixed in all.
- All tool tips under analyze work properly except one (assign temp labels in K-means)
- Chevron animation function is now defined in only one file
- `base_analyze.html` is reformatted
- Unused/unnecessary function and variables have been removed
- `scripts_analyze.js` meets ES Lint standards
- Multiple main functions in `scripts_base.js` are combined into one
- Grey words culling option has been removed from the front end to match the back end


Weiqi97:
As we discussed in #743, [import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) are now supported in newer version of Chrome and Firefox. I made a `utility.js` and put a couple functions that were commonly shared by most tools under analyze in it. I tested it on both Firefox and Chrome. It seems everything works just fine. But I need more vote on if we should keep things this way.

chantisnake:
Good choice adopting import and export. That is going to make our life much easier. Also, you need to import jQuery and other libraries.

Use `import * as $ from 'jquery'` to import jQuery. Or `import {$} from jquery`

mathewleblanc:
I tried using `import * as $ from 'jquery'` and I get the same error message as before:

SyntaxError: import declarations may only appear at top level of a module

I put the import statement at the very top of scripts_analyze.js.

scottkleinman:
Should it be this?

```javascript
import { $ } from jquery
```

mathewleblanc:
I just tried that too and I'm still getting the same error message.

scottkleinman:
After a little more reading, I wouldn't have posted that. What about some of the solutions [here](https://stackoverflow.com/questions/34338411/how-to-import-jquery-using-es6-syntax)?

Weiqi97:
I think we will work on importing other libraries in the future or at least on another branch..

PR #747, opened 2018-06-11
manage page: bug fix

phuens:


phuens:
the 'delete selected' on the right click was a space off compared to the others. I just put a space and fixed it.

Issue #681, opened 2018-01-22
jQuery 2.4.4 security vulnerabilities


scottkleinman:
jQuery should be updated to v3.0.0 to avoid security vulnerabilities described [here](https://snyk.io/test/npm/jquery/2.2.4?severity=high&severity=medium&severity=low). In the past, certain Lexos features have not worked with jQuery 3.0.0. I am not sure if this is still a problem (issues would only be detectable with extensive testing). The link provided gives some simple remedies which would be easier to fix before upgrading the jQuery version.

chantisnake:
in 3.0.0 the `complete`, `error` and `success` callback is removed, and I am pretty sure we might be using it in some of our functionality.

scottkleinman:
Removed or deprecated? If it is removed completely, that may be responsible for the errors I encountered in earlier tests. I wish I had noted at the time where these issues occurred. But this is a start.

scottkleinman:
A quick test shows that `complete`, `error`, and `success` do in fact work with the latest version of jQuery. They should probably still be changed in Lexos, but they are probably not responsible for any Lexos functions failing with jQuery 3.x.

chantisnake:
I suggest create a branch and change jquery to 3.0.0 right now.

mleblanc321:
@mathewleblanc and @danny98m, after talking with @Weiqi97  -- will you do this (i believe matt has  finished a trial)

mathewleblanc:
I looked up how to upgrade to jQuery 3.x [here](https://jquery.com/upgrade-guide/3.0/#browser-support).
It seems that we must add `<script src="https://code.jquery.com/jquery-3.0.0.js"></script>
<script src="https://code.jquery.com/jquery-migrate-3.0.1.js"></script>` to each page in order to upgrade. The first script is the new version of jQuery and the second is a plugin that identifies compatibility issues via the browser console. One of the main problems so far is that `.click(function(){})` has been deprecated in 3.x. We must now use `.on('click', function(){})` instead. So far I am not aware of any other errors except for `.tooltip()` and `.popover()` in scripts_base.js. How should I proceed? I have been changing the click functions as I go, should I continue to do this? I do not yet know how this will impact the Ajax.

scottkleinman:
[...]

PR #749, opened 2018-06-12
Scrub Refactor

mathewleblanc:
# Scrub Refactor

- Reformatted HTML and JS
- Changed vars to const and let
- Added JS doc to functions
- Broke up long functions
- Two script tags in html now combined into one
- Changed `getElement` to jQuery `$(selector)`
- Replaced string concatenation with template literals

I definitely need some help with the rest of this, mostly the stuff in the script tag in scrub.html. I tried moving it all to scripts_scrub.js but that did not work. I also tried to put the chevron rotate animation into a separate function as I have before but this does not seem to work either and I am not sure why. Because this animation is used on almost every page we may want to move it from base_analyze, scripts_scrub and anywhere else it is used and put it in scripts_base.


lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,309 @@
...
+}
+
+/**
+ * Send ajax request to do scrubbing
+ * @param {string} action - preview or apply
+ * @returns {void} - returns nothing
+ */
+function doScrubbing (action) {
+  if ($('#num_active_files').val() === '0') {
+    $('#error-modal').modal()
```
chantisnake: Probably add an error message?

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,309 @@
...
+  // Flask route seems to be the easiest method.
+  window.location = '/downloadScrubbing'
+}
+
+/**
+ * Send ajax request to do scrubbing
+ * @param {string} action - preview or apply
+ * @returns {void} - returns nothing
+ */
+function doScrubbing (action) {
```
chantisnake: I don't think this function does scrubbing.

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,309 @@
...
+    error: function (jqXHR, textStatus, errorThrown) {
+      $('#error-modal-message').html('Lexos could not apply the scrubbing actions.')
+      $('#error-modal').modal()
+      console.log('bad: ' + textStatus + ': ' + errorThrown)
     }
+  }).done(function (response) {
+    response = JSON.parse(response)
+    $('#preview-body').empty()
+    $.each(response['data'], function () {
+      const fileName = $(this)[1]
```
chantisnake: Avoid using `this`, and index

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,309 @@
...
+      const fileContents = $(this)[3]
+      const fieldSet = $('<fieldSet></fieldSet>')
+      // CSS truncates the document name
+      fieldSet.append('<legend class="has-tooltip" style="color:#999; width:90%;margin: auto; white-space: nowrap; overflow: hidden; text-overflow: ellipsis;">' + fileName + '</legend>')
+      fieldSet.append('<div class="filecontents">' + fileContents + '</div>') // Keep this with no whitespace!
```
chantisnake: Why don't you merge these three lines together and use template literal?

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,309 @@
...
+}
+
+/**
+ * Function to handle the chevron drop down button rotate animation by toggling
+ * the class so the appropriate CSS applies.
+ * @returns {void} - returns nothing
+ */
+function rotateChevron () {
+  $(this).find('span').toggleClass('down')
+  $(this).next().collapse('toggle')
```
chantisnake: What does `this` means? I think you can just pass that in as a parameter?

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,309 @@
...
+}
+
+/**
+ * Clone file labels and do ajax request
+ * @returns {void} - returns nothing
+ */
+function buttonFileLabelsFunction () {
+  // swfileselect, lemfileselect, consfileselect, scfileselect
+  const fileType = $(this).attr('id').replace('bttnlabel', '')
+  const usingCache = $('#usecache' + fileType).attr('disabled') !== 'disabled'
```
chantisnake: Template literal instead of adding strings.

lexos/templates/scrub.html
```diff
@@ -2,355 +2,537 @@
...
+              href="https://cdn.datatables.net/u/bs/jszip-2.5.0,pdfmake-0.1.18,dt-1.10.12,b-1.2.1,b-html5-1.2.1,b-print-1.2.1,fc-3.2.2,fh-3.1.2,se-1.2.0/datatables.min.css"/>
+        <script type="text/javascript"
+                src="https://cdn.datatables.net/u/bs/jszip-2.5.0,pdfmake-0.1.18,dt-1.10.12,b-1.2.1,b-html5-1.2.1,b-print-1.2.1,fc-3.2.2,fh-3.1.2,se-1.2.0/datatables.min.js"></script>
+        <script type="text/javascript"
+                src="{{ url_for('static', filename='node_modules/datatables.net-bs/js/dataTables.bootstrap.js') }}?ver={{ version }}"></script>
+    {% endif %}
+
+    <script type="text/javascript"
+            src="{{ url_for('static', filename='js/scripts_scrub.js') }}?ver={{ version }}"></script>
+    <script>
```
chantisnake: Avoid scripts in HTML.

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,309 @@
...
-    if ($('#aposhyph')[0].style.cssText == 'display: none;') {
-      $('#aposhyph').fadeIn(timeToToggle)
-    } else {
-      $('#aposhyph').fadeOut(timeToToggle)
-      $('#aposhyph')[0].style.cssText == 'display: none;'
+  // Do Ajax
+  $.ajax({
+    type: 'POST',
+    url: '/removeUploadLabels',
+    data: $(this).text().toString(),
```
chantisnake: I still don't know what is `this`...

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,309 @@
...
+        style: 'width:100px;margin:50px auto;z-index:1000;'
+      }).appendTo('#xmlModalBody')
+
+      $('#xmlModalStatus').append('<img src="/static/images/loading_icon.svg?ver=2.5" alt="Loading..."/>')
+    },
+    success: function (response) {
+      const j = JSON.parse(response)
+      const t = '<table id="tagTable" class="table table-condensed table-striped table-bordered"></table>'
+      $('#xmlModalBody').append(t)
+      const select = `<select id="allTags" style="margin-top:3px;margin-right:5px;"> \
```
chantisnake: You don't need line continuation symbol

lexos/templates/scrub.html
```diff
@@ -2,355 +2,537 @@
...
+      $(function () {
+        $('#save-button').on('click', function (e) {
+          e.preventDefault()
+          let tagDict = {}
+          const a = $(this).closest('form').serializeArray()
```
chantisnake: `a` probably is not a good enough variable name

lexos/static/js/scripts_scrub.js
mathewleblanc: What do you mean? There is already an error message in error-modal, this allows it to display.

lexos/static/js/scripts_scrub.js
mathewleblanc: `this` is the legend which contains a chevron drop down button that the user clicks. I'm not sure that it can just be passed as this click function is tied to these legends by class name, and passing that as an argument instead of using `this` will cause all of the drop downs to open and chevrons to rotate when only one is clicked.

lexos/static/js/scripts_scrub.js
mathewleblanc: I am actually not sure what `this` is in this context. The script in the HTML is poorly documented and its purpose is not immediately apparent.

lexos/static/js/scripts_scrub.js
mathewleblanc: I'm not sure if there is a good way to rewrite this section without using `this` and index. This section gets the file names and previews for each active file and adds them to the DOM. This information is stored in an array once the JSON is parsed so I am not sure how else we can do this without using either `this` or index.

lexos/templates/scrub.html
mathewleblanc: I am not completely sure what this script does. I removed it and have done some testing and everything seems to still we working properly. I cannot find many of the elements that this script refers to on the page.

lexos/static/js/scripts_scrub.js
mathewleblanc: I do not believe this ajax call is necessary. It does not do anything with the response. I just removed it and tested and it seems to have no impact.

lexos/static/js/scripts_scrub.js
chantisnake: I think you can just pass the `currentTarget` into this function.

Read this: https://api.jquery.com/event.currentTarget/

Also, please remove every single `this` that is in this code.

lexos/static/js/scripts_scrub.js
chantisnake: ``#usecache${fileType}``

lexos/static/js/scripts_scrub.js
mathewleblanc: In this case if `function (i, val)` is used, `i` is of course the index of each value but `val` is another array. How can I access specific indices of `val` without using `[1]` or `[3]`? I'm assuming that we do not want to use another `$.each()` within the one that is already there.

lexos/static/js/scripts_scrub.js
mathewleblanc: The only example on that page uses an anonymous function and passes `event` as a parameter and is then able to do `event.currentTarget`. How do I do the same with a normal function?

scottkleinman: What is the justification for removing `$(this)`? Are there places where it generates scoping issues?

lexos/static/js/scripts_scrub.js
chantisnake: That seems like a legacy problem of the back end. Typically in this case, we don't use number index, but use name index. Instead of `[1]`, we usually use stuff like `['filename']` for better readability.

In this case you can keep the number indexing.

lexos/static/js/scripts_scrub.js
chantisnake: The justification for removing `this` is to create robust functions. A function should does what is promise to do wherever it is invoked. If you use `this` in a function it is forced to be invoked in a specific scope. It creates robustness issue (we cannot reuse the function in other context), readability issue (people have to understand where the function is invoked to understand what the function does) and possible bugs (people might invoked the function in other context and hope it would 'handle the chevron drop down button rotate animation'). 

This is considered a good practice to avoid `this` outside of the context of a class see this eslint rule: https://eslint.org/docs/rules/no-invalid-this

scottkleinman: Hmmm... I need to evaluate this when I have more time. jQuery `$(this)` is not the same as Javascript `this`, so I am not sure that it shares the same scoping issues.

chantisnake: It probably does, since it depends on the environment of the outer scope.

lexos/templates/scrub.html
chantisnake: Hum, talk to Weiqi about this.

lexos/static/js/scripts_scrub.js
chantisnake: You need to explicitly add an error message here, in case the message is changed in the future. In fact, in just under 20 lines below this line of code, there is a line of code that modifies the error message.

lexos/static/js/scripts_scrub.js
chantisnake: I think this may be some bad practice on the backend. Please investigate on what /removeUploadLabels page does.

lexos/static/js/scripts_scrub.js
chantisnake:
```js
function rotateChevronOfLegend (legend) { 
  $(lengend).find('span').toggleClass('down') 
  $(lengend).next().collapse('toggle') 
}
```

when this is invoked:
```
$(stuff).on("click", function(e){rotateChevronOfLegend(e.currentTarget)})
```

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,292 @@
...
-  })
-
-  if ($("input[name='haveGutenberg']")) {
+  if ($('input[name=\'haveGutenberg\']')) {
     $('#gutenberg-modal').modal()
   }
   $('#actions').addClass('actions-scrub')
 
-  $('.has-chevron').on('click', function () {
+  $('.has-chevron').click(function () {
```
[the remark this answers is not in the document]
mathewleblanc: I am pretty sure that we decided [here](https://github.com/WheatonCS/Lexos/issues/681#issuecomment-391846858) to use `.click()`.

lexos/static/js/scripts_scrub.js
mathewleblanc: Ah, I see, this makes much more sense now. This actually fixed a problem I was having with the rotate animation.

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,292 @@
...
-      // check number of files
-      if ($('#num_active_files').val() <  1) {
-          // display error
-          $('#error-modal-message').html(msg)
-          $('#error-modal').modal()
-      }
-  })
-
-  if ($("input[name='haveGutenberg']")) {
+  if ($('input[name=\'haveGutenberg\']')) {
```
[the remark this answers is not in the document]
mathewleblanc: I actually didn't notice this change when I made it. PyCharm automatically changed it when I used their shortcut to reformat the file to our guidelines. I will change this back.

lexos/static/js/scripts_scrub.js
```diff
@@ -1,267 +1,311 @@
...
         $('#xmlModalStatus').remove()
-        if (j["selected-options"] != "multiple") {
-          $("#allTags option[value='"+j["selected-options"]+"']").prop("selected", true)
-        }
-        var value = $('#myselect option:selected').val()
-        var text = $('#myselect option:selected').text()
+        $('#allTags option[value=\'' + selection + '\']').prop('selected', true)
       },
       error: function (jqXHR, textStatus, errorThrown) {
         console.log('Error: ' + errorThrown)
```
Weiqi97: Maybe 
```javascript
`Error: ${errorThrown}`
```
Some other places could use template literal too.

Weiqi97:
Please add the Eslint checking for this script  to both Travis and Appveyor.

mathewleblanc:
Turns out scrubbing tags did not work in this branch. I just fixed this issue but I had leave a lot of function declarations inside the main function in order for it to work. It still meets ES lint standards, but I am still waiting to see if it will pass the other checks on Github.

PR #754, opened 2018-06-15
Capitalized Statistics block title so its uniform with the rest.

danny98m:
For whatever reason statistics was the only block title not capitalized so I changed it because it was irritating me.


Weiqi97:
My bad. Thanks.

PR #756, opened 2018-06-15
Feature/content analysis test suite

Krissac:
updated test suite example


Weiqi97:
@Landaluce If you have time, could you please take a look at this PR and give it a try to update it? Thank you very much.

chantisnake:
The phrase definition seems weird to me, since it hard codes ascii spaces and newlines etc, which is typically not that good of a idea.

chantisnake:
The phrase definition seems weird to me, since it hard codes ascii spaces and newlines etc, which is typically not that good of a idea.

I don't know if these are really necessary.

Weiqi97:
We addressed some issues presented here in #875 instead.

PR #758, opened 2018-06-17
K-Means Model

Weiqi97:
## Refactored K-Means tool to model
- Added a new 3D scatter plot visualization method.
- Voronoi visualization is done by plotly now.
- Created model for K-Means tool, updated view and front-end (Ajax call is used), old version of K-Means is removed.


lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,455 @@
...
+
+        # Fill the pandas data frame.
+        result_table["Cluster #"] = [index + 1 for index in k_means_index]
+        result_table["Document"] = labels
+        result_table["X-Coordinate"] = reduced_data[:, 0]
+        result_table["Y-Coordinate"] = reduced_data[:, 1]
+
+        return result_table
+
+    def _get_3d_frame(self, k_means_index: List[int]) -> pd.DataFrame:
```
chantisnake: Why is this function not covered by the unit test?

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,455 @@
...
+        result_table = pd.DataFrame(columns=["Cluster #",
+                                             "Document",
+                                             "X-Coordinate",
+                                             "Y-Coordinate"])
+
+        # Fill the pandas data frame.
+        result_table["Cluster #"] = [index + 1 for index in k_means_index]
+        result_table["Document"] = labels
+        result_table["X-Coordinate"] = reduced_data[:, 0]
+        result_table["Y-Coordinate"] = reduced_data[:, 1]
```
chantisnake: I guess a better way would be 
```python
result_table = pd.DataFrame(data={
    "Cluster #": [index + 1 for index in k_means_index],
    "Document": labels
    "X-Coordinate": reduced_data[:, 0]
    "Y-Coordinate": reduced_data[:, 1]
})
```

lexos/models/kmeans_model.py
chantisnake: Why is this function not covered by unit test?

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,455 @@
...
+                                             "X-Coordinate",
+                                             "Y-Coordinate",
+                                             "Z-Coordinate"])
+
+        # Fill the pandas data frame.
+        result_table["Cluster #"] = [index + 1 for index in k_means_index]
+        result_table["Document"] = labels
+        result_table["X-Coordinate"] = reduced_data[:, 0]
+        result_table["Y-Coordinate"] = reduced_data[:, 1]
+        result_table["Z-Coordinate"] = reduced_data[:, 2]
```
chantisnake: same thing as above

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,455 @@
...
+
+        :return: A plotly object hat has been converted to HTML format string.
+        """
+        # Get kMeans analyze result and unpack it.
+        k_means = self._get_k_means()
+        reduced_data = self._get_reduced_data()
+        k_means_index = k_means.fit_predict(reduced_data)
+
+        # This is important, such that plot and table results are consistent.
+        sorted_k_means_unique_index = sorted(set(k_means_index))
```
chantisnake: Do you need the `set`? are you trying to only take the unique term? maybe you can use `np.unique` and then `np.sort`?

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,455 @@
...
+        cluster_labels = [labels[np.where(k_means_index == index)]
+                          for index in sorted_k_means_unique_index]
+
+        # Get a list of lists of file coordinates based on the cluster result.
+        cluster_values = [reduced_data[np.where(k_means_index == index)]
+                          for index in sorted_k_means_unique_index]
+
+        # Get a list of centroid results based on the cluster result.
+        centroid_values = [np.mean(cluster, axis=0, dtype="float_")
+                           for cluster in cluster_values]
```
chantisnake: I think these are three parallel arrays, remember we conceal parallel arrays because they are bug prone. You can turn them into one panda data frame or two panda series.

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,455 @@
...
+        # This method could be updated once plotly better support polygons.
+        voronoi_regions = [go.Heatmap(x=x_mesh_grid[0][:len(z_value)],
+                                      y=x_mesh_grid[0][:len(z_value)],
+                                      z=z_value,
+                                      hoverinfo="skip",
+                                      showscale=False,
+                                      colorscale='YlGnBu')]
+
+        # Pick a color for following scatter plots.
+        color = cl.scales["10"]["qual"]["Paired"]
```
chantisnake: This function is awfully long (L165-L292, way over the 30-line limit for functions). Is there a way to split this function into several functions?

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,455 @@
...
+            go.Scatter(
+                x=[x_value[index]
+                   for index, group_index in enumerate(k_means_index)
+                   if group_index == group_number],
+                y=[y_value[index]
+                   for index, group_index in enumerate(k_means_index)
+                   if group_index == group_number],
+                text=[labels[index]
+                      for index, group_index in enumerate(k_means_index)
+                      if group_index == group_number],
```
chantisnake: These are awfully complicated, how about this:
```python
data = [
    go.Scatter(
        x=[x_value[index]], 
        y=[y_value[index]],
        text=[labels[index]]
    )
    for index, group_index in enumerate(set(k_means_index))  # I don't understand why this has to be have `set`, why not `np.unique`
]
```

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,455 @@
...
+                   if group_index == group_number],
+                y=[y_value[index]
+                   for index, group_index in enumerate(k_means_index)
+                   if group_index == group_number],
+                z=[z_value[index]
+                   for index, group_index in enumerate(k_means_index)
+                   if group_index == group_number],
+                text=[labels[index]
+                      for index, group_index in enumerate(k_means_index)
+                      if group_index == group_number],
```
chantisnake: same thing as above

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,455 @@
...
+
+    plot: go.Figure
+    table: pd.DataFrame
+
+
+class KMeansResult(NamedTuple):
+    """A typed tuple to hold processed k-means results."""
+
+    plot: str
+    table: str
```
chantisnake: Give these a alias for readability.
```python
PlotlyHTMLPlot = str
HTMLTable = str

class KMeansResult(NamedTuple):
    """A typed tuple to hold processed k-means results."""

    plot: PlotlyHTMLPlot 
    table: HTMLTable 
```

Weiqi97:
@chantisnake Thanks for reviewing. I know that the 2D scatter and 3D scatter are not covered by unit test. I meant to hear some advice back first.. And yes the reason I use ```set``` is that I need unique items. I will change them to ```np.unique```.

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,475 @@
...
+        x_low_bound = x_min - (x_max - x_min) / 5
+        x_up_bound = x_max + (x_max - x_min) / 5
+        # Increase 200 will make lines smoother.
+        x_step = (x_up_bound - x_low_bound) / 200
+
+        # Find min, max for y and then calculate bounds and step.
+        y_min, y_max = y_value.min(), y_value.max()
+        y_low_bound = y_min - (y_max - y_min) / 5
+        y_up_bound = y_max + (y_max - y_min) / 5
+        y_step = (y_up_bound - y_low_bound) / 200
```
chantisnake: I think you can create a helper function for these line, since you are repeating yourself here.

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,475 @@
...
+            else MatrixModel().get_id_temp_label_map()
+
+    @property
+    def _k_means_front_end_option(self) -> KMeansOption:
+        """:return: a typed tuple that holds the k-means front end option."""
+        return self._test_front_end_option \
+            if self._test_front_end_option is not None \
+            else KMeansReceiver().options_from_front_end()
+
+    def _get_reduced_data(self) -> np.ndarray:
```
chantisnake: If you have time I think you need to think about refactoring this method, since this method either returns a 2D array or a 3D array, which is two different data type.

chantisnake: If you don't have time, leave it like this. This is probably fine.

lexos/models/kmeans_model.py
chantisnake: If you don't have time, this is also fine.

lexos/models/kmeans_model.py
```diff
@@ -0,0 +1,475 @@
...
+        )
+
+        # Set the layout of the plot.
+        layout = go.Layout(
+            title="K-Means Voronoi Result",
+            xaxis=go.XAxis(title='x-axis', showline=False),
+            yaxis=go.YAxis(title='y-axis', showline=False),
+            hovermode="closest")
+
+        # noinspection PyTypeChecker
```
chantisnake: Is this line in use? if not please remove it.

lexos/static/js/scripts_kmeans.js
```diff
@@ -1,172 +1,139 @@
...
+        tableDiv.children().DataTable(getDataTableConfiguration())
+        // Insert the plot result.
+        $('#KMeans-plot').html(response['plot'])
+      })
+    .fail(
+      function (jqXHR, textStatus, errorThrown) {
+        // If fail hide the loading icon.
+        $('#status-analyze').css({'visibility': 'hidden'})
+        console.log('textStatus: ' + textStatus)
+        console.log('errorThrown: ' + errorThrown)
```
chantisnake: Maybe template literal here.

lexos/static/js/scripts_kmeans.js
```diff
@@ -1,172 +1,139 @@
...
+
+/**
+ * Return the desired data table configuration.
+ * @returns {{paging: boolean, scrollY: number, scrollCollapse: boolean, dom: string, buttons: string[], columnDefs: *[]}} data table configuration.
+ */
+function getDataTableConfiguration () {
+  const hideColumns =
+    $('#vizMethod').val() === '3DScatter' ? [2, 3, 4] : [2, 3]
+
+  return {
```
chantisnake: I don't think this is a good idea. imaging a typed system, when you type `{pag<tab>}`, you will not get any intellisense, since the code analyzer don't know what object you are creating.

But if you type `table.DataTable({pag<tab>})`, you will get intellisense here, because the code analyzer knows what type `.DataTable()` method takes.

Therefore I recommend you to change to the following code:
```js
function makeDataTable(table) {
  const stuff = stuff
  table.DataTable({paging: false, other stuff..})
}
```

lexos/templates/kmeans.html
```diff
@@ -2,443 +2,147 @@
...
-    -o-transition:      all 0.25s ease;
-    transition:         all 0.25s ease;
-}*/
-</style>
-
+    <!-- Lexos K Means Script -->
+    <script type="text/javascript"
+            src="{{ url_for('static', filename='js/scripts_kmeans.js') }}?ver={{ version }}"></script>
+    <!-- Latest plotly js -->
+    <script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
```
chantisnake: specify a version

lexos/templates/kmeans.html
```diff
@@ -2,443 +2,147 @@
...
+    <!-- Latest compiled and minimized DataTables CSS -->
+    <link rel="stylesheet" type="text/css"
+          href="https://cdn.datatables.net/v/bs/jszip-2.5.0/pdfmake-0.1.32/dt-1.10.16/b-1.5.1/b-colvis-1.5.1/b-html5-1.5.1/b-print-1.5.1/fc-3.2.4/fh-3.1.3/r-2.2.1/datatables.min.css"/>
+    <!-- Latest compiled and minimized DataTables JS -->
+    <script type="text/javascript"
+            src="https://cdn.datatables.net/v/bs/jszip-2.5.0/pdfmake-0.1.32/dt-1.10.16/b-1.5.1/b-colvis-1.5.1/b-html5-1.5.1/b-print-1.5.1/fc-3.2.4/fh-3.1.3/r-2.2.1/datatables.min.js"></script>
+    <!-- Datable plug in to truncate the columns. -->
+    <script type="text/javascript"
+            src="https://cdn.datatables.net/plug-ins/1.10.16/dataRender/ellipsis.js"></script>
```
chantisnake: I think you need to include npm version of these packages. See other pages for example

lexos/templates/kmeans.html
chantisnake: specify a version, see `package.json`

chantisnake:
I have made couple comments, please read them before merging.

lexos/templates/kmeans.html
Weiqi97: Yeah.. I think I may create a separate PR for this.. Since the same line exists in ```similarity.html``` and ```statistics.html```. Same thing for plotly package.

lexos/models/kmeans_model.py
Weiqi97: I wanted to move on to something else.. So I left a ```TODO``` for this.

PR #760, opened 2018-06-18
Refactor topword model

Weiqi97:
### Topword now is done by ajax call.
- Better visual for class divisions.
- Remove unused options.


scottkleinman:
Before this is merged, I'd recommend one more change on the front end. When you try to compare all documents without assigning any classes, you get the following message: "Only one class found, if desired to compare between class: please go to the [Manage page](http://127.0.0.1:5000/manage) to assign class labels to documents". This should be changed to "No classes have been assigned or all active documents belong to a single class. If you wish to compare documents of different classes, you must first assign them class labels on the [Manage screen](http://127.0.0.1:5000/manage)."

Weiqi97:
@scottkleinman The message was updated. Thanks for noticing that.

Issue #444, opened 2016-07-16
Re-implement TF-IDF in Normalize options


scottkleinman:
We need a fresh implementation of TF-IDF weighting. See issue #437 for the background.

chantisnake:
no body really understand how TF-IDF works, that is the problem...

scottkleinman:
OK, I'll look into it as soon as I can.

scottkleinman:
[This](http://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.TfidfTransformer.html#sklearn.feature_extraction.text.TfidfTransformer) helps a lot. Simply using `tf * idf` as the formula will match the standard accounts known to humanists. This would be equivalent to the `norm: None` setting. We can find out if there is any demand for fine-tuning the normalisation parameters at a later date. The real question is whether we should offer (or even impose) the `smooth_idf` setting (which is the default). This adds 1 to to document frequencies, "as if an extra document was seen containing every term in the collection exactly once" (I'm quoting the sklearn manual here.) "The effect of this is that terms with zero idf, i.e. that occur in all documents of a training set, will not be entirely ignored." If we can decide whether this refinement is worthwhile for Lexos, the implementation should be relatively easy.

scottkleinman:
I just did some testing using both `sklearn.feature_extraction.text.TfidfTransformer` and `sklearn.feature_extraction.text.TfidfVectorizer`. If I set `Norm=None` I get a matrix exactly the same as the raw counts. If I use any other normalisation setting, I get an empty matrix. We may have to play with these functions outside of Lexos to figure out where the problem is.

In the meantime, I changed the commented out html in `base_analyze.html` a little bit to make it easier to test when we get around to it.

scottkleinman:
After discussion, we've decided to leave this for Lexos 3.5 and label this issue an enhancement.

Issue #437, opened 2016-07-14
Weighted Counts in Tokenize don't all work 


scottkleinman:
In the Weighted Counts (TF-IDF) options, only Norm:None works (not Euclidean and Manhattan Distance). This is an issue in Lexos 2.5, as well. I don't really understand what Norm:None means (and how it is different from not using weighting), so I'm unwilling to attempt a fix. We should decide what we want to offer for these options.

scottkleinman:
This is a near duplicate of issue #269. Closing that one. The consensus there was to remove Euclidean and Manhattan distances. However, I am still not sure what Norm:None means.

mleblanc321:
in base_analyze.html, i've removed the TF-IDF options from the UI;  we have issues with this code but no time to work on it now, plus i know of no one actually using or requesting this option

scottkleinman:
Is Norm:None still in there? I would use TF-IDF if I could figure out what exactly Norm:None is doing.

scottkleinman:
I think we should re-implement TF-IDF soon but start fresh. I'm going to close this issue and start a new one.

Issue #269, opened 2016-01-08
TF-IDF producing empty matrices?


scottkleinman:
I just had a query from a user about TF-IDF in the Tokenize screen. When he selected it, he got back a "No data is available" message. When I try it, I get a table in which every cell is 0 is I use Euclidean or Manhattan distance. Norm:None gives me a result. I had noticed this before and assumed that it was the nature of my data, but we might want to check whether something is going on with the Euclidean and Manhattan functions, or at least get an idea of what our expected output should be for the three options. I'm not even sure what Norm:None means. If there is no bug, we should do a quick write up in *In the Margins*.

mleblanc321:
TF-IDF never got a thorough review; for now, i'd say we
should remove it

m



On Fri, Jan 8, 2016 at 5:01 PM, scottkleinman <notifications@github.com>
wrote:

> I just had a query from a user about TF-IDF in the Tokenize screen When he
> selected it, he got back a "No data is available" message When I try it, I
> get a table in which every cell is 0 is I use Euclidean or Manhattan
> distance Norm:None gives me a result I had noticed this before and assumed
> that it was the nature of my data, but we might want to check whether
> something is going on with the Euclidean and Manhattan functions, or at
> least get an idea of what our expected output should be for the three
> options I'm not even sure what Norm:None means If there is no bug, we
> should do a quick write up in *In the Margins*
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/WheatonCS/Lexos/issues/269>.
>



-- 
-----------------------------------------------------------------------------
Mark D. LeBlanc, Ph.D.
Professor of Computer Science
Co-Chair, Department of Math & Computer Science
Wheaton College, Norton, MA 02766
508.286.3970

http://cs.wheatoncollege.edu/mleblanc
-----------------------------------------------------------------------------

scottkleinman:
That makes things easier for short-term development, but I think this is a desired functionality, so we should probably not put it off too long.

chantisnake:
Do we need to remove that now?

scottkleinman:
Oh, good catch. I don't know if this has been addressed. The last time I looked at this only the Norm: None setting produced a non-empty matrix.

scottkleinman:
Transferred to issue #437.

PR #762, opened 2018-06-19
Fix TF-IDF return type

Weiqi97:
In ```matrix_model.py```, the TF-IDF method returns a ```csv_matrix``` type instead of a ```pandas.Dataframe```. Thus using this normalize method causes problems. Since in all other models that access ```matrix_model.py```, we assume the DTM returned is ```pandas.Dataframe``` type. This is a quick fix for the issue. I did a couple trial runs and it seems to be working okay.


Weiqi97:
From issue #437 and #269, @scottkleinman do you may remember which file was causing problem? Or it didn't work at all? I can look into this a bit more as we wanted a fresh re-implementation on this from #444. Thanks in advance.

PR #770, opened 2018-06-22
Rolling Window Analysis Model

Weiqi97:
### A brand new rolling window analysis implementation created by @chantisnake 

- Rolling window now is in models.
- The graph is now done by plotly and it is more interactive than before.
- It allows multiple milestones now.


Weiqi97:
Unfortunately, I wasn't able to figure out a way to make milestones plot interactive. So it doesn't have hover over. And once it's plotted in the graph, it can not be hidden. One possible way I thought of was to plot each milestone line as dense straight scatter plots. But it will be computationally expensive...

Weiqi97:
And I'm still working on more unit tests.

scottkleinman:
I had hoped that Plotly would have fixed the lack of a hover event on shapes by now, but no such luck. The [latest enquiry](https://community.plot.ly/t/possible-to-add-tooltips-hovertext-to-shapes/10476) yielded the advice to add `opacity: 0 `scatter markers point along your shapes. I wonder if this is actually necessary for a milestone, which is just a vertical line. Could we simply add one at `x=n, y=0` and then trigger the effect if the mouse passes within a couple of pixels of `n` anywhere on the y-axis? That will require some calculation, but it might be better than plotting dense points all along the y-axis.

My other suggestion, which doesn't necessarily address the same issue would be to have a toggle button somewhere to toggle the shapes on and off (perhaps by changing their `opacity`).

Still, I think what you've got is a major improvement.

chantisnake:
I haven't look at the code. I have only one thing to add, that is to group the hack with the hack and group the good code with the good code.

In the case of plotly fixing this issue, we can simple remove all the hacks and not to change the major codebase.

Also I understand the interactablity is hard to add plotly given their current logic, so they probably will not fix it soon.

lexos/models/rolling_windows_model.py
```diff
@@ -0,0 +1,574 @@
...
+        :return: An array of strings, each element is a window.
+        """
+        # Number of items in the input list.
+        num_item = len(input_list)
+
+        # Get the total number of windows.
+        num_window = num_item - window_size + 1
+
+        # Get the rolling list, should be a array of str.
+        return np.array([
```
Weiqi97: @chantisnake After some testing around, I found this line is extremely slow when it comes to the case that user specifies a large window size. Assume we have a file that is 500KB (which isn't large at all, apparently we support file size up to 250MB) and we specified the window size to be 10000, I think we will end up saving something around `500KB * 10000`. And in fact, this often causes Lexos to crash with exit code `247` when windows size is big. I didn't find much details about this exiting code but I assume it is memory related. I have another branch that I tried to get rid of this line, so we don't need to save all windows. It won't crash at that point but it still take a really long time to process. Any suggestions on improving this will be appreciated.

chantisnake: Hum, I am guessing this is caused by the join function being slow. don't really know how to fix this.

chantisnake: I have experimented with  `join` method they are super fast. 
```python
In [22]: import uuid  # use to generate random string

In [23]: a = [str(uuid.uuid4()) for _ in range(10000)] # 10000 random strings

In [24]: a[:5]  # these are long random strings
Out[24]:
['752779e9-3260-46c7-9072-c3604d45711c',
 '8e5526e8-74bb-4fc6-83dc-33dc8f709fb0',
 '4484acff-758d-4b9d-a72f-eb45e1c59e31',
 '143a168e-4df5-421d-a8b9-db03b1b969f8',
 'dfdec4f2-a0ce-45eb-ab2d-10691ac82a0d']

In [25]: %time b = "".join(a)  # time the concatenation
Wall time: 0 ns

In [26]: b[0:500]  # to preview that they are really joined
Out[26]: '752779e9-3260-46c7-9072-c3604d45711c8e5526e8-74bb-4fc6-83dc-33dc8f709fb04484acff-758d-4b9d-a72f-eb45e1c59e31143a168e-4df5-421d-a8b9-db03b1b969f8dfdec4f2-a0ce-45eb-ab2d-10691ac82a0d5951c65c-6fc8-4329-b79a-6c0007406fb00feb4e73-251b-4a55-8455-1ebed1d2834fa627f844-0f00-4d5f-9725-5342f6dda46d58064618-789d-42ae-bccb-a181796bc95152c774ad-572a-47c6-b380-cf9482c89ba0ca2b91b6-6032-4999-9f98-90e8a1c34c4934450420-a4c6-4388-878f-d32992d593b91dc18257-0971-488d-9bb6-0d23df80daa5577f0422-ec1c-4e7c-88d9-cfe6cb4a'
```

There are no way they can cause lag. I don't know the reason for this.

lexos/models/rolling_windows_model.py
```diff
@@ -0,0 +1,574 @@
...
+                    x1=mile_stone,
+                    y0=0,
+                    y1=y_max,
+                    line=dict(
+                        color=self._get_mile_stone_color(index=index),
+                        width=1
+                    )
+                )
+                for index, key in enumerate(mile_stones)
+                for mile_stone in mile_stones[key]]
```
chantisnake:
```python
for index, (milestone_str, milestone_x_coord) in enumerate(mile_stone_dict.itmes)
```

lexos/models/rolling_windows_model.py
```diff
@@ -0,0 +1,574 @@
...
+        ]
+
+    def _add_milestone(self, result_plot: List[go.Scattergl]) -> go.Figure:
+        """Add milestone to the existing plot.
+
+        :param result_plot: List of existing scatter rolling window plot.
+        :return: A plotly figure object.
+        """
+        # Get all mile stones.
+        mile_stones = self._find_mile_stone_windows_indexes_in_all_windows(
```
chantisnake: change this to `milestones_dict`

chantisnake:
As for the interactivity of the milestone. I think we have discussed this in slack @scottkleinman , Do you still have the code snippet to interact with shape?

chantisnake:
Here is a link that might be relevant to displaying milestone https://stackoverflow.com/questions/42152660/how-to-add-a-name-or-label-to-shapes-in-plotly

scottkleinman:
@chantisnake Thanks for pulling out that old code. I would have had quite a time finding it.

PR #771, opened 2018-06-22
Change dendrogram result appearance

Weiqi97:
- Make the result graph more flat, so it could be displayed one screen.
- Add a little space on right side of the graph, so the most right line won't be overlapping with the border of the graph.


Weiqi97:
This solution will cause Dendrogram to crash when the orientation is not bottom.. I will find another possible solution.

scottkleinman:
I am writing up some materials that might help.

PR #777, opened 2018-06-25
Tokenizer Model

Weiqi97:
## Tokenizer is in model now. 
- Tokenizer is fully integrated with `matrix_model.py`
- Changed the appearance of the button to toggle columns on and off.


Weiqi97:
This change is also on [lexos3.wheatoncollege.edu](http://lexos3.wheatoncollege.edu).

chantisnake:
I am so glad this is finally done.

chantisnake:
However, we do need to handle sorting, and slicing the dtm, I don't know where these are handled.

lexos/models/tokenizer_model.py
```diff
@@ -0,0 +1,252 @@
...
+        dtm_head = file_row_dtm_soup.find("thead")
+
+        # Form the total frame and set column name.
+        total_fame = pd.DataFrame(
+            columns=np.round(file_row_dtm.sum(axis=0), 3))
+        total_fame.columns.name = "Total"
+
+        # Convert the total frame to beautiful soup object.
+        total_soup = BeautifulSoup(total_fame.to_html(classes="table"),
+                                   "html.parser")
```
chantisnake: I don't quite get why do you have to convert the frame to an html? html is much harder to manipulate than dataframe, since the data structure can be way to robust.

Weiqi97: I was trying to put two more rows in the table header : `total` and `average`. So that these two rows are fixed when page changes. But I guess there should be a better way.

chantisnake:
I imagined this api will be like this. The frontend send the backend a request of this form:
```typescript
{sortby: Alphabetical | Str, 
 sortOrder: SmallToLarge | LargeToSmall, 
 pageNum: Int, 
 pageSize: Int}
```

Obviously json cannot have type `Alphabetical | str` (this means either sort alphabetically or sort by a word), we can have the following hack:

```typescript
{sortByAlphabetical : bool, 
 sortByWord: str | null,
 sortOrder: 'SmallToLarge' | 'LargeToSmall', 
 pageNum: Int, 
 pageSize: Int}
```

Therefore if we want to sort the word alphabetically (which is the default) the `sortByAlphabetical` will be true, if we want to sort by the count of a particular word, we use `sortByWord`.

In this way, we can only send the frontend a slice of the DTM, that will reduce the commutation needed on the frontend and the reduce the size of data that is transmitted through the web.

Weiqi97:
I will close this PR. At this moment, tokenizer works fine but it does not have the server side processing enabled. The branch will be kept and I will put a bit more work on it.

chantisnake:
Although I am happy that this pull request is made. I think this model still needs some major improvements.

Weiqi, have you tested the speed of this app? In terms of sorting and generating the transposed dtm. I think this tool should at least handle 5 mobi dicks.

chantisnake:
I think we should **at least** match the speed of previous tokenizor tool.

Weiqi97:
@chantisnake I know I have not enabled server side processing and I've been trying to find solutions. I'm closing this since at this moment we really should not merge it.

PR #781, opened 2018-06-25
Bug-Fix/upload

phuens:
Added the loading icon when you upload a file.


phuens:
@Weiqi97 The load spinner nows spins in Google chrome and Firefox. We used font awesome for the spinner icon rather than uploading our own image. Everything looks good to me.

PR #786, opened 2018-06-25
Added autoscroll functionality to analyze tools

danny98m:
When results are generated the screen will scroll to them. 
This works in: 
* K-Means
* Dendrogram
* Sim Query
* Topwords
* Statistics


Weiqi97:
Nice job. Is it possible that this function could be put into to `utility.js`? It could take in the div id as a parameter.

danny98m:
@Weiqi97 Thank You!

PR #805, opened 2018-06-29
Added Browser Detection

mathewleblanc:
# Added Browser Detection

This adds primitive browser detection to warn the user if they are using an unsupported browser. This solution is hard coded until we can figure out how to use a library for browser detection as a long term solution.


chantisnake:
This is not an urgent feature and the code is not up to the standard. We need to use a library that is well-tested and accepted by the community. Current code is just a hack. 

I will not agree to merge this pull request.

Weiqi97:
I am closing this since we will seek for a library that does this...

PR #811, opened 2018-07-28
Bugfix/dendro appearance

Weiqi97:
### Fix dendrogram plotly graph
- Removed right and top orientation. 
- Added a padding so that most right leaf will not touch the border of the plot. (Note: The padding now will not go away when you zoom in and zoom out.)
- Dynamically extend plot margins so that long label will not get cut off.

### Below are two usage cases.
- Left orientation
screenshot from 2018-07-28 16-29-00
- Bottom orientation
screenshot from 2018-07-28 16-29-59


mleblanc321:
hmm, i'm testing DAZ and segment labels are being truncated/chopped? (using lexos3)

PR #812, opened 2018-08-02
Rolling Windows Enhancement

Weiqi97:
### Enhance rolling window model
- Speed up generating the rolling windows.
- Add a button to allow user download rolling window results as a CSV file.


Weiqi97:
This is a naive testing that compares the new implementation of function `_get_rolling_window_from_list` with the old implementation. (Sorry the code may not be easy to read...)
```python
def get_next_window(window, last_word, next_word):
    window = window.replace(last_word, "", 1)
    return "".join([window, next_word])

# Test suite
word_list = [str(i) for i in range(100000)]

# First method.
program_starts = time.time()
window_list_one = ["".join(word_list[start: start + 10000]) for start, _ in enumerate(word_list[:100000-10000+1])]
program_ends = time.time()
print("The first method takes: %f seconds" % (program_ends - program_starts))

# Second method.
program_starts = time.time()
window = "".join(word_list[:10000])
window_list = [window]
for index, next_word in enumerate(word_list[10000:]):
    window = get_next_window(window, word_list[index], next_word)
    window_list.append(window)
program_ends = time.time()
print("The second method takes: %f seconds" % (program_ends - program_starts))

if window_list_one == window_list:
    print ("The two are equal")
```
And the results are:
```python
The first method takes: 7.503632 seconds
The second method takes: 0.554603 seconds
The two are equal
```
The second method is the newer method and it is usually more than 10 times faster than the old method.

Weiqi97:
From the discussion with Mark, we believed that the users usually will not have too many milestones. (Possibly not over 1000 milestones.) After some testing, I believe draw all mile stones as straight lines does not effect the speed much. 
So, now users can hide milestones on click.

PR #817, opened 2018-08-18
Bootstrap Consensus Tree

Weiqi97:
The long waited Bootstrap Consensus Tree analysis is here! (After a while I finally figured out a good way to send a picture to flask... And unit test will come soon...)
Now it is available on [Lexos3](lexos3.wheatoncollege.edu) under analysis -> clustering.
Let's see a screen shot of a sample run here.. It looks like a dendrogram set to left orientation. 
screenshot_20180818_001443
I think a circular tree will certainly look better, but ```ete3``` has failed working with flask. I'm now seeking alternative solutions with ```networkx``` while trying to communicate with ete3 maintainers for possible workarounds. But I suppose this change can be a future improvement, or even better, plotly may support a circular tree in the future.
#### Here is a couple things that I couldn't quite make the decision myself or I have concerns about.
- Data replacement related issues.
  - Should we allow users decide how much data they want from the DTM per iteration? Right now, we grab 80% of the data from the DTM each iteration, and users cannot change this number.
  - Should we enable an option that allow [replacement](https://www.ma.utexas.edu/users/parker/sampling/repl.htm) when picking data from the DTM per iteration? Right now we are sampling without replacement.
- In the screen shot, you can find that when a label is too long (Roughly when the label is over 35 characters), it will be truncated. If 35 characters is not enough, we can change the margin size but it will shrink the tree branches.
- Do we really need the Y-Axis? Or should we name it something other than ```taxa```?
- And.. I believe speed will be a concern.. Especially when users have a lot of files and when they want to run hundreds of iterations. Now the lowest number of iterations they can put down is 20 while the highest is 200. I'm not sure if those numbers make sense. And should we alert them that the process may take long when they choose to run hundred of iterations?
###### Any suggestions are welcome, thank you for reviewing.


scottkleinman:
Great work (though we'll need an explanation of how it is working&mdash;I can't tell what is being iterated from the UI).

I'm surprised that you're having trouble with ete3. It should be possible to save the radial dendrogram as an SVG to session folder and then load it into the UI. I think we still do that with png images of dendrograms in Lexos 3.1, so you could have a look at the code for that. You can also save ete's Newick data and generate a graph using Javascript libraries. You can use a [pure d3 implementation](https://bl.ocks.org/git-ashish/3aa81521f96e48198c80b4e2742bb6bc), an older library like [jsPhyloSVG](http://www.jsphylosvg.com/), or a newer one like [TnT Tree](http://tntvis.github.io/tnt.tree/). It actually is possible to graph a radial tree in Plotly, as you can see from [this example](https://plot.ly/~empet/14676/bcl-2-gene-family-tree/#plot). But I have no idea how you generate the plot points.

Still, we've needed this for a very long time. Thanks for pursuing it.

Weiqi97:
@scottkleinman Thank you for the suggestions! I believe any one of the JavaScript libraries that supports newick format will work great. (I'll try look into them and do some testing). The program works first finding the newick format of the consensus tree and then display it with matplotlib, so we should be able to alter it easily.
(For the ete3 solution, I was able to get it completely working with jupyter notebook. But when I put it on the server, it only works when you click on generating the result the first time. And when you try to generate the result again, ete3 renders an empty picture. However if you start a new session, it will work again for only the first time. I looked at the log and found ```WARNING: QApplication was not created in the main() thread``` and ```QObject::startTimer: Timers can only be used with threads started with QThread```. I tried to turn off ```flask``` multi-thread setting, but it did not work. I think the error came from PyQt library. I tried to chase it but no luck. (Though I think it might me doing something wrong since no one else has ever asked a similar question about ete3...)

Weiqi97:
I'm hoping to work on ```in the margins``` with Mark together for this tool and along the way I'll add more explanations to the UI as well. Thank you for catching it!

PR #825, opened 2018-08-26
Add an error page when scikit-bio is missing

Weiqi97:
@scottkleinman suggested this approach and I think this is the best way to remain only one release. 
So when `bct_view` detects that the scikit-bio library is missing, it renders a different page with some options user can choose from. This way, even if scikit-bio is missing, users can still use the rest of Lexos.

### DIFF
--- a/lexos/templates/bct_analysis_import_error.html
+++ b/lexos/templates/bct_analysis_import_error.html
@@ -18,17 +18,18 @@
                 detect this library on your machine.
             </li>
             <li class="list-group-item-text">
-                If you with to use Lexos without Bootstrap Consensus Trees,
+                If you with to use Lexos without Bootstrap Consensus Tree,
                 you may ignore this message. The other Lexos tools will run
                 fine.
             </li>
             <li class="list-group-item-text">
-                if you with to use Bootstrap Consensus Trees tool without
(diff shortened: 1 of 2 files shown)


scottkleinman:
There is an error in the template file. It should read "If you want to have this tool running on your localhost, you can install **scikit-bio**" instead of "If you want to have this tool running on your localhost, you can install **scikit-learn**".

I would also recommend a couple of improvements to the language. Something like this:

The Bootstrap Consensus Trees tool requires the <code>scikit-bio</code> Python library. Lexos was unable to detect this library on your machine.

- If you wish to use Lexos without Bootstrap Consensus Trees, you may ignore this message. The other Lexos tools will run fine.
- If you wish to use Bootstrap Consensus Trees tool without installing <code>scikit-bio</code> on our machine, you can visit our <a href="http://lexos.wheatoncollege.edu">server</a>.
- If you want to run Bootstrap Consensus Trees on your own computer, follow the instructions below.

> Mac/Linux users: Open a terminal and run <code>pip install scikit-bio</code>.
> Windows users: Please follow the instructions on our <a href="https://github.com/WheatonCS/Lexos/wiki/How-to-Install-scikit-bio-on-Windows">Wiki</a>.

PR #840, opened 2019-05-14
Update dev_requirements.txt

Weiqi97:


chantisnake:
Have tried using that to create exe, did not work well. Hope you can succeed.

Weiqi97:
@chantisnake Thanks! I think the package is bit outdated now. I will try to use PyInstaller as an alternative. [PyFladesk](https://github.com/smoqadam/PyFladesk) looks like a possible way for us to pack the entire flask app with the frontend as executable. I will spend some time poking around.

Weiqi97:
cx_freeze removed in #843, since it causes error and crashes appveyor.

chantisnake:
It will be great if you can make it work.
the problem I have encountered is that numerical library like Numpy and Matplotlib have a really hard time to package.
Because they used lots of c code, and they can be painfully system specific.

PR #842, opened 2019-05-14
Bugfix/close tooltips

### LINKED ISSUE #841: Tooltips not closing
Some specific tooltips behave differently than others; they do not disappear when mouse un-hovers the ?, only when the tip itself is subsequently hovered and un-hovered.
e.g. - "Distance Metric" on pages "Hierarchical clustering", "Bootstrap consensus tree"
- "Weighted Counts (TF-IDF)" on pages "Hierarchical clustering", "K-means clustering", "Bootstrap consensus tree"
- "Sample each iteration" on page "Bootstrap consensus tree"

jacksonjreed:
fixed #841


jlobes:
great catch - 
I also noticed this bug in Analyze -> Similarity Query -> Select Comparison Document in case you didn't catch that one

jacksonjreed:
you're correct--links no longer clickable. I still think the current behavior is less than ideal (is there an element that disappears unless you mouse over it immediately?), but I'll close this pull request.

Issue #837, opened 2019-05-14
Tokenizer: convert to pandas dataframe?


mleblanc321:
Tokenize. Perhaps experiment with assembling the data in a pandas dataframe, converting the head to html using `pandas.to_html()`, and sending that to the front end. It might be more efficient and actually make the Tokenizer table display work.

jacksonjreed:
@Weiqi97 can this one be closed with the tokenizer rewrite?

chantisnake:
It is better if you add the link to the pr you are referring to.
Thank you.

jacksonjreed:
that would be #873

Weiqi97:
I believe #873 resolves this issue.

PR #846, opened 2019-05-16
Tokenizer rewrite

Myles-Trevino:
This is a rewrite of the Tokenizer page in response to issue #837. The page no longer throws errors on switching pages of the DTM table and does not freeze on large datasets. All functionality is retained except the display of the averages and visualization of the "Documents as Rows, Terms as Columns" data orientation option.

The data orientation was not retained as it appears to be impossible to display many columns efficiently with the current tables library as columns cannot be paginated while rows can. The CSV download however still reflects the data orientation option change. To signify this, "(Applied On Download)" text was added next to the data orientation option.

The display of averages was not retained as calculating the averages noticeably reduces performance when navigating the table and can be done in various other software once the CSV is downloaded.


lexos/views/tokenizer_view.py
```diff
@@ -1,579 +1,197 @@
...
+# This is a flask blueprint. It helps us to manage groups of views. See here
+# for more detail:
 # http://exploreflask.com/en/latest/blueprints.html
 # http://flask.pocoo.org/docs/0.12/blueprints/
-tokenizer_blueprint = Blueprint('tokenizer', __name__)
+tokenizer_blueprint = Blueprint("tokenizer", __name__)
+
+
+def get_session_dtm_options(default_orientation: bool =
+                            True) -> Dict[str, object]:
```
chantisnake: When are we starting to use dict now? I remember we are using named tuples.

lexos/views/tokenizer_view.py
```diff
@@ -1,579 +1,197 @@
...
-                row = "<tr>"
-                for s in l:
-                    row += "<td>" + str(s) + "</td>"
-                row += "</tr>"
-                rows += row
-        # Catch instances where there is no active document (triggers the error
-        # modal)
+
+            # Get the DTM as a list of tuples
+            dtm = get_dtm_matrix(get_session_dtm_options(), file_manager)
```
chantisnake: why are we not using matrix model here?

chantisnake:
I think this is a great improvement. I am really glad it is done, but I am not sure this is the best we can do.

lexos/views/tokenizer_view.py
chantisnake: Also, don't forget to put documentation on every function.

Weiqi97:
@chantisnake Thank you for reviewing and I do agree with you. I communicated with @Myles-Trevino earlier. I have an implementation on the Tokenizer model from last summer which did not include the `ServerSide` functionality. I am currently working on it and Myles's implementation has been very helpful.

lexos/views/tokenizer_view.py
Weiqi97: My bad. I have not talked about the MVC structure yet. I will try to give the team a quick introduction on it tomorrow.

Weiqi97:
Most issues discussed here are addressed in #873

Issue #806, opened 2018-06-29
Active documents icon triggers when you try to upload invalid file types


scottkleinman:
If you try to upload an invalid file type (e.g. `.json`) in Upload, you, the upload is prevented, but the Active Documents icon appears, and the number of active documents shown increments if you try again.

phuens:
@scottkleinman I will look into it.

jacksonjreed:
working on this--
there seems to be a larger associated issue: if you navigate to the upload page and upload a new batch of documents, the active documents count updates to display only the number of docs you *just* uploaded

jacksonjreed:
the code to initialize is accessing #counter from the html, which is set to 0 when loading the page...
the incrementation issue is fixed but waiting to push this until I figure out how to properly access active docs count

jacksonjreed:
never mind, I'm pushing a branch to fix this specific issue...
BUT this code needs to be cleaned up more, the counter still definitely doesn't work as intended. I'll open another issue,

PR #861, opened 2019-05-20
active docs count when uploading wrong filetype

jacksonjreed:
fix for #806


chantisnake:
Maybe a better way is to "always" query the backend for number of active document, to make sure the backend and the frontend are consistent?

jacksonjreed:
> Maybe a better way is to "always" query the backend for number of active document, to make sure the backend and the frontend are consistent?

hoping to do something like that in fixing #862--this is more of a temporary fix to close the original issue.

chantisnake:
I thought we have talked about the option to always query the backend on how many files do you have?

jacksonjreed:
@chantisnake my plan was to do another pr to add that behavior more consistently across a few pages--however with the front end rewrite most of this stuff will get replace anyway.

chantisnake:
Good! I am glad

Issue #865, opened 2019-05-20
wordcloud word count display issues


jacksonjreed:
currently clicking in the word cloud and clicking (not necessarily adjusting) the angle at the bottom both redraw the graph, but in different ways.
clicking in the word cloud produces a cloud that is properly scaled, but (above a certain threshold) draws only a fraction of the number of words the user asks for:
![1](https://user-images.githubusercontent.com/50586789/58052696-86f84480-7b23-11e9-902f-6e1eaa8a7eb1.png)
clicking on the angle seems to produce the correct number of words, but not properly scaled so there is a lot of overlap:
![2](https://user-images.githubusercontent.com/50586789/58052735-9d9e9b80-7b23-11e9-81d5-e8f7627418de.png)
these first two are graphs of 200 words. in both cases a higher "number of words" input does increase the number of words in the chart, but the first method still displays many fewer words than the second. see the charts for 5000 words:
![3](https://user-images.githubusercontent.com/50586789/58052877-fc641500-7b23-11e9-9428-7c8b8861b95a.png)
![4](https://user-images.githubusercontent.com/50586789/58052884-00903280-7b24-11e9-9c22-6440f357a96e.png)

there might be a connection #370 which asks that the graphs be re-generated automatically when _any_ option is changed; hopefully investigating that one will help with this one.

PR #868, opened 2019-05-22
Bugfix/wordcloud

jacksonjreed:
this is a temporary fix for the messy behavior referenced in #865, pending a larger overhaul using newer version of d3cloud. word counts will still be wrong above ~75. regenerating graph from angle adjust has been removed so clicking in the graph is the only way now.


jacksonjreed:
@mleblanc321 We could merge this temporarily but it is made obsolete by the front end rewrite,

PR #873, opened 2019-05-22
Tokenizer Model

Weiqi97:
### Converted Tokenizer into the MVC structure
- Added `script_tokenizer.js` to EsLint check.
- Removed deprecated DTM functions.
- Removed option to allow users view files as rows. (They can still download in that orientation.)


.travis.yml
```diff
@@ -37,6 +37,7 @@ matrix:
         - ./node_modules/.bin/eslint ./js/scripts_similarity.js
         - ./node_modules/.bin/eslint ./js/scripts_dendrogram.js
         - ./node_modules/.bin/eslint ./js/scripts_rwanalysis.js
+        - ./node_modules/.bin/eslint ./js/scripts_tokenizer.js
```
chantisnake: Why not use a regex to match all the scripts?

lexos/models/tokenizer_model.py
```diff
@@ -0,0 +1,190 @@
...
+    """A typed tuple that holds all the tokenizer test options."""
+
+    token_type_str: str
+    doc_term_matrix: pd.DataFrame
+    front_end_option: TokenizerOption
+    id_temp_label_map: IdTempLabelMap
+
+
+class TokenizerModel(BaseModel):
+    """This is the class to generate statistics of the input file.
```
chantisnake: I think you forget to change this

lexos/models/tokenizer_model.py
```diff
@@ -0,0 +1,190 @@
...
+    def _token_type_str(self) -> str:
+        """:return: A string that represents the token type used."""
+        if self._test_token_type_str is not None:
+            return self._test_token_type_str
+        else:
+            # Get dtm front end options.
+            dtm_options = MatrixReceiver().options_from_front_end()
+            # Get the correct current type.
+            token_type = dtm_options.token_option.token_type
+            return "Terms" if token_type == "word" else "Characters"
```
chantisnake: What are the acceptable values that token type can hold?

.travis.yml
Weiqi97: I think we haven't made all the files to pass EsLint yet...

chantisnake: Ohohoh, yeah, sure.

chantisnake: Nope, no need to sorry, things need to be done one at a time. I understand.

chantisnake:
Looks okay to my pity human eyes.

PR #875, opened 2019-05-23
Bugfix/content analysis/add file

cgnobrega:
### Content Analysis Page
- Improved PEP8 documentation, return type hinting, and the usage of keyword parameters
- **Active Docs icon** now displays 
- Loading icon still does not display. *This could be solved by someone with more JS experience/knowledge*


lexos/views/content_analysis_view.py
```diff
@@ -53,14 +59,17 @@ def content_analysis():
...
-                               itm="content-analysis",
-                               formula=formula)
+        return render_template(
+            'contentanalysis.html',
+            dictionary_labels=dict_labels,
+            active_dictionaries=active_dicts,
+            toggle_all_value=toggle_all_value,
+            itm="content-analysis",
+            formula=formula,
+            labels=id_label_map,
```
Weiqi97: I'm not really familiar with this tool... Could you please check if those two values `labels` and `numActiveDocs` are used in `contentanalysis.html`?

cgnobrega: In `base.html` (`contentanalysis.html` extends `base.html`) the value `labels` is not used, but `numActiveDocs` is, this is what allows the active docs icon to display on the Content Analysis page of the web tool. So line 69 is actually not necessary given that it is not used to display the active docs icon.

PR #876, opened 2019-05-23
Bugfix/viz_fixes

jlobes:
### Fixes to BubbleViz UX:
* Added new 'Save as SVG' button
* Changed 'Save as PNG' Button to just download the png
  * Removed code where instead it would redirect to new page with the PNG pasted there
* Both buttons only appear after 'Get Graph' has been clicked and results are loaded


jlobes:
# New Updates
### Newer Fixes to BubbleViz UX:
* `mouseover` on a bubble now gives you the entire word along with the count - only in Firefox
* Fixed caching issue where reloading the graph would reset the custom options on the page
* Cured most of the eslint errors thrown, some items still not defined
### Older Fixes to BubbleViz UX:
* Added new 'Save as SVG' button
* Changed 'Save as PNG' Button to just download the png
  * Removed code where instead it would redirect to new page with the PNG pasted there
* Both buttons only appear after 'Get Graph' has been clicked and results are loaded

jlobes:
> Maybe use string interpretation?
What do you mean?

Weiqi97:
If I am interpreting Cheng correctly, I think he meant `string interpolation`. So `string literals` which allow the embedded expressions, similar to the `f-string` in Python. In JS, we have the following syntax:
```javascript
const variable = "Mark"
const string_one = "My name is: " + variable
const string_two = `My name is: ${variable}`
```
Notice that the string literal is enclosed by ` not ', and do not forget the dollar sign.
(My fault. I forgot to mention this at the beginning.)

lexos/static/js/scripts_viz.js
```diff
@@ -1,5 +1,11 @@
+import d3 from './Lexos/lexos/static/node_modules/d3'
```
Weiqi97: I'm not sure if this `import` statement would work. Could you please double check?

PR #878, opened 2019-05-23
Only trigger codecov in Travis

Weiqi97:
I think we may do not need to trigger codecov from both CIs? @chantisnake It will be great if I could get your opinion on it.


chantisnake:
Reasonable.

PR #883, opened 2019-05-29
scrubber unit tests--enabling some tests and speedup

jacksonjreed:
some changes to the scrubber code and unit tests to enable three unit tests that were previously disabled and speedup many of the tests


chantisnake:
Do not leave code in comments

PR #887, opened 2019-05-31
Bugfix/scrubber whitespace handling

mlimoges:
Updated the handling of removing whitespace to include all unicode whitespace, also updated the test to match.

Unfortunately, this had to be done as a static list.


mlimoges:
This one should be good to go. Sorry about not keeping it up to date.

PR #891, opened 2019-06-03
Bugfix/scrubber digits decimal detection

jlobes:
# Decimal Detection in `scrubber.py`
## Fixes to `scrubber.py` and `test_scrubber.py`:
* changed `get_remove_digits_map` to incorporate a regex which also accounts for decimals numbers when **Remove Digits** is ticked, also had to change around the test suite function in `test_scrubber.py` that accounted for it.


lexos/processors/prepare/scrubber.py
```diff
@@ -461,32 +461,25 @@ def get_remove_punctuation_map(
...
-        remove_digit_map = dict.fromkeys(
-            [i for i in range(sys.maxunicode)
-             if unicodedata.category(chr(i)).startswith('N')])
+    # Using "d" to represent a digit, character, "."
+    # to represent any unicode character used to indicate
+    # a decimal number, and "***" to represent any sequence of
+    # any digits, this pattern will match:
+    # 1) ***
+    # 2) ***.***
+    pattern = re.compile(r"(\d)|((?<=\d)[\u0027|\u002C|\u002E|\u00B7"
```
mleblanc321: we should capture as much as possible, so i think we want (\d**+**) [add + for runs of at least one (unicode) digit ... as well at the end of the regex \d**+**)

chantisnake:
I think removal map would not work in this case, because we are matching complicated regex, not just removing fixed strings, (like a dash)

Also, please use verbose regex, this regex is not that readable.

jlobes:
## Final Changes:
* changed the digits regex in `scrubber.py` to optionally accept signed numbers
* edited the unit test in `test_scrubber.py` accordingly
* documented the `get_remove_digits` function to properly describe the new way that we handle digits
* updated the test suite accordingly, adding two new files: `new_regex_jun19.txt` and `new_regex_jun19_Results.txt`

Issue #893, opened 2019-06-05
UI Fixes


cgnobrega:
### Overall 
- [x] sorry, but i (mdl) don't like the new icon;  it is kinda scary looking (in fact, it looks like a snake, not a dragon now); where is filled-in color icon?
- [x] Lexos (then space)  /Tool   we should be consistent, either space on both sides of slash(/) or not
- [x] "Help" should have a constant highlight color
- [x] Check boxes need check mark
- [x] Radio buttons need dot in center
- [x] Full footer (we want NEH logo; MIT License?)
- [x] Git hub test suite needs update to reflect directions for new UI
- [x] the buttons need more "pop" (hard to locate): standardize color/size/position
- [x] tool tips needs a better image, crisper
- [x] (perhaps only an issue on Firefox) numerical text fields can't be entirely selected
- [x] .lexos workspace files can be uploaded, but are unusable. attempting to load scrub gives a file not found error, other pages simply fail to properly load and show the error bar.
- [x] Tool names of each page should be in header bar
- [x] visualizations don't display single chars
- [x] walk-thru icon
- [x] manage files by clicking active docs (low priority)
- [x] Analyze tools- PNG and SVG downloads error "invalid resolution"

### Upload
- [x] hmm, i don't like that upload has lost the prime real estate;  sure, some (not many) like to enter URLs, but the space granted (one-half of opening screen) is *not* indicative of our users imho; the drag-files-here (marching ants) box should cover the entire screen, left to right;  (consider some other small area to upload URLs)
- [x] Upload list - document boxes are a bit big and bulky, looked better when narrow
- [x] Make marching-ants ---- border more pronounced
- [x] Save workspace.lexos works, downloads file; upload of .lexos works, 
- [x] When uploading multiple files, if one in the middle of the upload list is empty, the rest don't get uploaded
- [x] as of latest version "active documents" counter displays "NaN" upon uploading files, but changes to the correct number after loading another page
- [x] when uploading multiple documents (e.g. use the four apostrophes test suite files), only one appears in the uploads list (though they all appear on the manage page)

### Manage
- [x] I miss "click to toggle" the Select/Deselect feature;  that is, if on, click and it gets deselected; (is this problematic to keep?)
- [x] Text fields for “Edit Name” and “Edit Class” are very small and don’t save input if you were to edit again
- [x] inconsistent text in the "excerpt" section (sometimes a few words, sometimes 3 lines, sometimes only 1 line)
[...]

PR #899, opened 2019-06-07
Feature/front end rewrite itm

cgnobrega:
### Help sections
- all help sections for all tools written (except the not yet migrated content analysis page)
- some sections contain descriptions for buttons that are yet to be function, but refer to issue #893 for these page specific functionality issues/errors 
- made the "active docs count"  link to the manage page (like in original)


chantisnake:
It is not good practice to use h3 tag for something that is not a header. You should you a class and css to set the style for that paragraph.

PR #902, opened 2019-06-08
Feature/circular bct

Landaluce:
Add radial graph to BCT


Weiqi97:
Since some code changed in the model, please update the unit test as well; at least remove the tests that are no longer relevant. Thanks!

chantisnake:
I think there may need some documentation on JavaScript code?

chantisnake:
Can't you use npm to install the new library you need? Why do you need to copy and paste it in the project?

chantisnake:
Also, there are about 400 lines of JavaScript code in the HTML file, with almost all single letter variable name without a single line of documentation.
I think we need to clean up these code.

mleblanc321:
circular dendograms are still on my list :)  .... but i think this code should be moved to [future work] list? thoughts

scottkleinman:
The best approach to this might now be to integrate the Lexos Python library, so I am closing this PR.

PR #914, opened 2019-06-13
Feature/front end rewrite stats page

maggieshafer:
Fixed quartiles to match the box plot
Added x and y-axes to box plot and scatter plot
changed the plots so that they show each point plotted
fixed bug fixes on the front end
added full footer
changed average terms to vocabulary density and added the term to the help page and glossary
added license to the about lexos page and how to cite us


lexos/models/stats_model.py
```diff
@@ -133,18 +134,32 @@ def _get_document_statistics_dataframe(self) -> pd.DataFrame:
...
-            file_stats[f"Total number of {self._token_type_str}"] / \
-            file_stats[f"Distinct number of {self._token_type_str}"]
+        file_stats[f"Vocabulary Density"] = \
+            file_stats[f"Distinct number of {self._token_type_str}"]/\
+            file_stats[f"Total number of {self._token_type_str}"]
+
 
         return file_stats
 
+    def __get_quartile__(self, index: float, arr: np.array) -> float:
```
Weiqi97: Even if it is a private function, it should be properly documented. This function might be a `staticmethod`.

lexos/models/stats_model.py
```diff
@@ -159,9 +174,20 @@ def get_corpus_stats(self) -> CorpusStats:
...
+            q1_index = 0
+            q3_index = 0
+        else:
+            q1_index = length * .25 + .5 - 1
+            q3_index = length * .75 + .5 - 1
+
         # Get the iqr of the file word counts.
-        first_quartile = file_sizes.quantile(0.25)
-        third_quartile = file_sizes.quantile(0.75)
+        first_quartile = self.__get_quartile__(index=q1_index, arr=arr)
```
Weiqi97: Does the performance of this function outrun numpy's performance? What was the reason to recreate the wheel?

chantisnake: I agree, also it seems confusing to me how they get q1_index, what is the purpose of the if statement, which set both q1_index and q3_index to zero?
Does numpy quantile not handle it correctly?

chantisnake:
For a PR that promised to rewrite stats front end, it has way too many changes that is irrelated to stats. There are changes to .idea files, changes to test suite.
I don't think these changes are desirable.

lexos/models/stats_model.py
maggieshafer: The numpy quantile calculates the quartiles in a different way than how plotly calculates them and the results are slightly different and Mark wanted them to be the same results. Setting both q1_index and q3_index to zero is for when only one document is put through stats. If they are not set to zero, q1_index is a negative number and it breaks.

chantisnake: Adding these explanation in the comments will be helpful for future developers.

lexos/models/stats_model.py
chantisnake: Also this function doesn't really get the `index` quartile of `arr`, I would recommend it actually gets the input quartile of the array (also, are you assuming the `arr` is sorted?)

PR #917, opened 2019-06-17
Feature/front end rewrite stats page

maggieshafer:
Added documentation to my quartile function


chantisnake:
I don't think the get_quartile function actually get the quartile of the array, but instead it gets the index of the array.

For example, getting the 50 quartile should return the mediean of the array.

maggieshafer:
I get the index of the array in get_corpus_stats and get_quartile gets the quartile of the array because it returns the value of the array at that index. I think you might be getting confused because if a quartile is in between two values it adds different weights to those values when calculating the actual quartile.

chantisnake:
For a function called `get_quartile`, it should get the quartile of a array.

That is
```python
__get_quartile__(0.75, [0, 2, 3, 1, 4])
```
Should return 3, because the 75% quartile of [0,2,3,1,4] is 3.

But your function did not do that.

PR #919, opened 2019-06-17
Bugfix/fix unit tests

WeaselPeasel:
The majority of the failing unit tests have been fixed. 16 are still left.


WeaselPeasel:
I just realized i left some stray print statements in the code, specifically in test_stats_model. I'll fix that, as well as try to maybe get a few more unit tests working.

PR #945, opened 2019-06-21
Front end rewrite/table sort

jacksonjreed:
enabled sorting for sim query table, stats table, and content analysis overview table
this code is kinda messy, maybe someone who understands it a little better than us can review it before it merges @Myles-Trevino @Weiqi97 @chantisnake @Landaluce


chantisnake:
You mean the code in this PR is kinda messy or the original code is kinda messy?

The change on simquery backend seems okay to me.
Telling me which part is messy might help with the review. Thanks!

chantisnake:
Also I am curious, we did not handle sorting in the previous version because it is handled by data table.
Are we migrating away from data table?

jacksonjreed:
I was mostly concerned with the content analysis page which received the biggest changes--the structure of the receiver was a little different from the other ones so we weren't sure whether we were using it correctly.

jacksonjreed:
not sure about the data table thing but the tables in this version definitely aren't sortable without a little extra work.

chantisnake:
It was @Landaluce 's creation. He should take a look.

chantisnake:
I would say probably you guys should investgate data table, it makes desplaying data easier.
It has many functionality and can save us lots of work.

PR #971, opened 2019-06-26
rwa milestones graph scaling

jacksonjreed:


jacksonjreed:
hmm the tokenizer col labels are also listed in the diff, although those changes were already merged.

PR #975, opened 2019-06-27
Front end rewrite/scrub tags options fixes

jacksonjreed:
- saved radio options now display when re-opening box
- replace text box enabled/disabled by radio options


Myles-Trevino:
Please use get_id() from utility.js rather than converting all the button names manually.

jacksonjreed:
I can't find that function anywhere but I agree there should be a better way to do this

PR #1024, opened 2020-06-10
Generalise fullscreen feature for all viz tools

scottkleinman:


scottkleinman:
The unit test is still looking for Early English HTML, so the whole build is failing. This not related to the changes in this branch. I am going to debug the unit test and then try to run it again.

PR #1029, opened 2020-06-10
Replace Consolidations feature with Pattern Replacement

scottkleinman:
This is a major change, replacing the old Consolidations feature with a new Pattern Replacement feature that implements a general pattern replacement function accepting regex patterns. It needs a review for UX and code efficiency. It probably also needs a unit test set up (volunteers?).


scottkleinman:
The frontend refactor of the last couple of days will make it very hard to merge this branch into master. I recommend just using this branch to test for UX and optimisation. If it works, the changes can be re-created in a new branch based on the latest version of master.

scottkleinman:
At this stage, the most likely follow-up for this PR is through the Lexos Python library, so I am closing it.

PR #1075, opened 2024-03-20
Fixed countVectorizer for scikit-learn version 1.3.2

JeffreyYStewart:
Updated scikit-learn to version 1.3.2 and updated associated code sections that used removed methods to use the newer versions.


Tozturk18 (approved):
Looks good! 

What is the change in definitions.py at line 26?
Is _WORD_REGEX_STR already unicode?

